PR #1, opened 2021-02-23
Use webpack dev server for faster development

R167:
Using `npm start` runs the webpack dev server. This has a few advantages:
- Runs in development mode so code compiles faster
- Automatically hosts a dev server on localhost (no more "go to `file:///User/wmd/.../index.html`).
- Better source map emission from typescript (type annotations are maintained)
- Hot reloading
- Uses HtmlWebpackPlugin to automatically inject the `webstart.js` entry point

This additionally significantly speeds up builds for incremental changes (e.g. <1s when changing a source file vs ~6s for the entire project).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,6 @@
+# ChocoPy WASM Compiler
+
+## Development
+
+To start in a local dev server, run `npm start`. This will also live reload
+any file changes you make.
--- a/index.html
+++ b/index.html
@@ -1,63 +1,62 @@
-<!doctype HTML>
+<!DOCTYPE html>
 <html>
-<head>
-    <meta charset="utf-8">
+  <head>
+    <meta charset="utf-8" />
     <script src="https://cdn.jsdelivr.net/gh/AssemblyScript/wabt.js@1.0.20/index.js"></script>
-    <script>const wabt = WabtModule;</script>
-    <script src="webstart.js"></script>
+    <script>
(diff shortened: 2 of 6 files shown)


jpolitz:
Thanks for sending this over. The whitespace diffs remind me that we wanted to get some kind of tslint rules in effect/project settings so we don't have the issue that everyone who opens a file causes a diff. Are the changes to the HTML file necessary/should the whitespace changes be a part of the commit?

In general, I have no objection to this at all, though, and appreciate it!

R167:
Whitespace in the html file here makes no difference. Realized part way though I have [Prettier](https://prettier.io) turned on for auto formatting code, etc. I would definitely advocate for prettier/eslint/tslint to be enabled.

As an example, the `.prettierrc` file I use and have found to be ~reasonable is:
```json
{
  "tabWidth": 2,
  "useTabs": false,
  "semi": true,
  "singleQuote": false,
  "printWidth": 100
}
```
(2 space indent, max line length 100, always append semi colons, and always use double quotes [esp b/c JSON])

I can go back and try to remove the white space changes for the time being though (might also add a README.md line about "use `npm start` for development").

R167:
@jpolitz Fixed the whitespace changes and opened #3 for a base eslint config.

jpolitz (approved):
LGTM

PR #2, opened 2021-02-23
CI best practices (use actions/cache for npm install cache)

R167:
Update the CI workflows to follow a couple best practices

- Use `npm ci`: This installs using _only_ `package-lock.json`, ensuring CI runs with the expected environment, doesn't upgrade packages, etc. More specifics available in the [npm docs](https://docs.npmjs.com/cli/v7/commands/npm-ci).
- Use `actions/cache` to save the `~/.npm` package cache. [docs](https://github.com/actions/cache/blob/main/examples.md#node---npm)
- Run distinct commands seperately (e.g. run an install phase **then** test phase. This can make debugging exactly which command failed slightly easier).

Additionally, here's an [example run](https://github.com/R167/chocopy-wasm-compiler/runs/1964134319?check_suite_focus=true) testing these changes.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,7 +2,7 @@
 
 name: CI
 
-# Controls when the action will run. 
+# Controls when the action will run.
 on:
   push:
     branches: [ main ]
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -23,9 +23,15 @@ jobs:
...
-      # Runs a set of commands using the runners shell
-      - name: Run and test
-        run: |
-          npm install
-          npm test
+      # Cache the npm install cache
+      - uses: actions/cache@v2
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
```
jpolitz, line 30: Is there documentation for where this incantation comes from?

jpolitz (approved):

.github/workflows/main.yml
R167: Yeah, copied directly from https://github.com/actions/cache/blob/main/examples.md#node---npm

ptliu:
Should make a note in README documenting committing your package-lock.json

R167:
> Should make a note in README documenting committing your package-lock.json

@ptliu I'm unclear on what you mean/point you're making here. Could you elaborate on the situation you're facing?

PR #3, opened 2021-02-23
Base eslint

R167:
Briefly [mentioned](https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/pull/1#issuecomment-784474950) in #1, this is a basic starting point for enabling linting using Prettier and ESLint. Rule set is the basic ones recommended by ESLint and typescript along with prettier.

This PR does **not** include any changed files other than config (since it would change ~every file).


jpolitz:
Do you understand the CI failure above? Seems like maybe this merge needs to come with some formatting updates.

R167:
@jpolitz im happy to do the formatting updates. wasn't sure if you wanted the formatting bundled with the config update though (because it makes the initial config review more obscured).

R167:
Applied the discussed prettier config and eslint to the codebase. Nearly all of the changes were systematic and straightforward.

jpolitz:
How much is this going to wreck the merges that folks are about to do tomorrow? Should we wait until those are in, then apply this once, then tell everyone to re-pull? That seems better.

R167:
@jpolitz I'm good to do that. Just not sure if there's really a "best time" to make this change since people will be working on it fairly consistently

The formatting changes themselves are easy enough to apply programmatically and probably could be merged successfully with git (i.e. prettier reformatting), it's just all the manual changes that can't be autofixed which are a pain.

I definitely get the point you're making about getting all the minor changes in and can make a quick sprint of applying the fixes ~friday afternoon. Maybe mention in class about adding linting and hold off on new bigger changes until friday evening (or something)? Just let me know!

R167:
I've updated this PR so it currently will only apply the ESLint linting rules and **not** the prettier reformatting. This should make it a lot easier to merge with the proposal submissions. After that, its trivial to enable prettier.

jpolitz (changes requested):
This is good, but I want to resist making too big a change all at once. People are going to have to _learn_ each new rule that's introduced, why it's different, etc. So I think we should drop some of the `var` restrictions that require a lot of rewriting for now.

I'd like to first focus on _formatting_, just whitespace/newlines, and making it possible and easy for folks to have their editor accommodate this. The rest I see causing a lot of pain.

jpolitz:
I didn't understand the degree to which these things were coupled in the initial commit, sorry. To say more – it seems like people have various whitespace settings in their VSCode setups that cause annoying and meaningless noisy merges. Is it easy to solve _just that_ problem?

If it's too hard to decouple/too much work to do that, maybe we have to live with the new rules. But I think this isn't quite solving a problem that we have.

Maybe we should have a separate PR that's _just_ `prettier`? I'm done merging everything significant for the ast, so it's a good moment to do `prettier` and then enforce it on folks.

R167:
> This is good, but I want to resist making too big a change all at once. People are going to have to _learn_ each new rule that's introduced, why it's different, etc. So I think we should drop some of the `var` restrictions that require a lot of rewriting for now.

I agree with the sentiment here, though I would also argue the `var` -> `let`/`const` are some of the more important changes to enforce given that `var`'s scoping semantics are unfortunate to say the least and can clear up some issues (esp. in case statements) where variables are being redefined.

> I didn't understand the degree to which these things were coupled in the initial commit, sorry. To say more – it seems like people have various whitespace settings in their VSCode setups that cause annoying and meaningless noisy merges. Is it easy to solve _just that_ problem?
> 
> If it's too hard to decouple/too much work to do that, maybe we have to live with the new rules. But I think this isn't quite solving a problem that we have.
> 
> Maybe we should have a separate PR that's _just_ `prettier`? I'm done merging everything significant for the ast, so it's a good moment to do `prettier` and then enforce it on folks.

I agree we could decouple a bit of this PR more if that's desirable. The structure of the rules somewhat outlines as follows
- `.editorconfig`: https://editorconfig.org common way that most editors understand to configure basic whitespace/indentation style
- `.prettierrc`: minimal overrides to prettier (in particular, there's ~no reason to limit lines to **only** 80 characters)
  - This config is actually entirely disabled right now. PR currently contains only the ESLint rule changes. This was done so merging back in main would be easier since prettier just reformats the entire file
- `.eslintrc`: definition of eslint rules. Most of the recommended ruleset are targeted at JS Best Practices and avoiding common mistakes
  - Currently most of the autoformatting rules are turned off because `prettier` usually will take care of these, however to make maintaining the branch easier, I have temporarily disabled prettier until we want to pull the trigger on merging this. Otherwise the diff is ~`changed most lines` and would be worse to review.

I think a reasonable path forward right now is turn off some of the more intrusive rules like `no-var`. For the most part, the code was good with most of the ruleset.

Plan:
- More minimal ESLint ruleset
- Get Review
- Turn on prettier (because aforementioned "changes every file" review issue)
- Confirm final review

Sound reasonable @jpolitz?

R167:
Ideally we also would have turned on the typescript `strictNullChecks: true` option, but that isn't really feasible without quite  a few changes (even though it would be really nice to avoid some bugs with non-exhaustive case-statements or returning AST nodes from type checking which are missing the type annotations)

.eslintrc
```diff
@@ -0,0 +1,47 @@
...
+  "root": true,
+  "parser": "@typescript-eslint/parser",
+  "parserOptions": {
+    "project": "./tsconfig.json"
+  },
+  "plugins": ["@typescript-eslint"],
+  "extends": [
+    "eslint:recommended",
+    "plugin:@typescript-eslint/recommended",
+    // "plugin:prettier/recommended",
```
R167, line 15: Note: To enable `prettier`, we need to uncomment this line and remove the `"requirePragma": true` from `.prettierrc`

R167:
Opened #38 which shows the current configuration with `prettier` applied on top of it. That branch is targeted at `main` and gives a full picture of what _all_ the formatting changes looks like.

jpolitz (commented):
I still wish this wasn't so large. I don't understand why we need new code in order to turn on formatting. But I'm approving so we can make progress.

jpolitz (approved):

R167:
> I still wish this wasn't so large. I don't understand why we need new code in order to turn on formatting. But I'm approving so we can make progress.

@jpolitz I just thought about it some more and the instances where code was changed should be a warning instead since the rules were more geared towards promoting safety in those instances than actually fixing the underlying problem. Additionally, the solutions aren't necessarily trivial, so it might be better suited as a linting warning anyways.

PR #4, opened 2021-02-24
add .toString() for Type

R167:
Currently parts of the compiler that must report arbitrary type errors typically do so through string interpolation, such as in field assignment

https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/blob/ed7fd8a05306e3140afd97d9ac8bc81bc96eba43/type-check.ts#L214

As it currently stands, this will print type `[object Object]` when it tries to display the type. I propose adding a prototype to `Type`s which implements `toString()` and resolves this issue. This issue was additionally present in type checking return statements where we would blindly cast to `any` and try to access the `name` field (leading to error messages of "expected return type `undefined`; got type `undefined`" for functions like `def r(): return 5`)

### DIFF
--- a/ast.ts
+++ b/ast.ts
@@ -1,11 +1,23 @@
 // import { TypeCheckError } from "./type-check";
 
+export class TaggedType<T extends string> {
+  tag: T;
+  constructor(tag: T) { this.tag = tag }
+  toString(): string { return this.tag }
+}
+
+export class ClassType extends TaggedType<"class"> {
+  name: string;
+  constructor(name: string) { super("class"); this.name = name }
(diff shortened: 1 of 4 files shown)


jpolitz:
This is another one that I don't want to force folks to deal with until after proposal review. I also think the type-checking group needs to see this – maybe mention some of them on Ed and point them to this issue to chime in?

R167:
That sounds reasonable. I was treating the inconsistent behaviour right now as a bug, but i'm also good to make mention of it on Ed for a relevant team. Maybe feels more like error messages though (since it's primarily intended to deal with how printing the types works)?

PR #5, opened 2021-02-24
Testing

ptliu:
Proposal for Testing team. Includes a change to package.json as well as a basic `npm run fuzz` command that currently runs unit tests, but will run autogenerated tests in the future.

Question: should we PR to main branch or testing branch first?


ptliu:
@edwjchen

jpolitz:
PR to main branch. We'll try to make it so if tests pass, things build.

Which reminds me, testing team – one additional task you could take on that would be superb would be (basic) Selenium tests that load the web page and check that e.g. running some simple programs produces the expected output. Any interest in trying that?

Full review to come.

edwjchen:
Hi Joe! 

I think we can take up adding in some end-to-end testing for the frontend as well. I have heard mixed reviews about using Selenium vs. Cypress for testing the frontend. Will look into both libraries a bit more and see which matches our use case more!

ptliu:
@jpolitz sounds good, looks like a lot of people are making PRs to their branch though and therefore not triggering PRs

jpolitz:
Once we merge this to `main`, it should trigger rebuilds on all the PRs that have `main` as a target. Is that sufficient for what you're concerned about?

ptliu:
I was more concerned about how, from looking at the PRs, they don't seem to target main, so they don't trigger CI. We don't expect to need to modify the CI pipeline with fuzzing, more so probably spinning up an AWS instance that fuzzes continuously. Might need CI modifications for cypress, but we'll cross that bridge when we get to it.

package.json
```diff
@@ -16,9 +17,11 @@
   "license": "ISC",
   "dependencies": {
     "@types/node": "^14.14.20",
+    "cypress": "^6.5.0",
     "lezer-python": "^0.13.1",
     "lezer-tree": "^0.13.0",
     "mochawesome": "^6.2.1",
+    "python-shell": "^2.0.3",
```
jpolitz, line 24: Why do you need python-shell?

jpolitz (approved):

package.json
edwjchen: Going to use python-shell in typescript to evaluate generated python programs! After generating a python program, we would evaluate the correctness of the program by comparing the output of the compiled python-to-wasm code (by running the compiled wasm code) and the generated python code (using python-shell).

Example use case of python-shell:
`
import {PythonShell} from 'python-shell';

PythonShell.runString('Insert Python Program As String Here', null, function (err) {
  if (err) throw err;
  console.log('finished');
});
`

PR #6, opened 2021-02-24
Project Proposals and Starting Points: Comprehensions

thecooltechguy:
This is our team's pull request for our proposal (we're working on comprehensions).

Our tiny change to the codebase is adding our new AST structures to `ast.ts`


ast.ts
```diff
@@ -37,6 +37,8 @@ export type Expr<A> =
   | {  a?: A, tag: "lookup", obj: Expr<A>, field: string }
   | {  a?: A, tag: "method-call", obj: Expr<A>, method: string, arguments: Array<Expr<A>> }
   | {  a?: A, tag: "construct", name: string }
+  | {  a?: A, tag: "comprehension", expr: Expr<A>, field: string, iter: Expr<A>, cond?: Expr<A> }
```
jpolitz, line 40: This seems like a good place to start. Good to see `block` here! I think given WASM's block structure it should be reasonable to codegen for, but will be interesting to see how that goes.

jpolitz (approved):

PR #7, opened 2021-02-25
Install directions clarification (and npm v6 usage)

R167:
@jpolitz Sanity checks to enforce using npm v6 mentioned during office hours. This is to prevent churn of `package-lock.json` due to the change in lock file format npm v7 introduced.


jpolitz (approved):
Looks great, thanks!

PR #9, opened 2021-02-25
Project Proposal for list group

erauyeung:
This is our project proposal and small codebase change for the Lists group.


ast.ts
```diff
@@ -25,6 +26,7 @@ export type Stmt<A> =
   | {  a?: A, tag: "while", cond: Expr<A>, body: Array<Stmt<A>> }
   | {  a?: A, tag: "pass" }
   | {  a?: A, tag: "field-assign", obj: Expr<A>, field: string, value: Expr<A> }
+  | {  a?: A, tag: "list_assign", list: Expr<A>, index: number, value: Expr<A> }
```
R167, line 29: 2 points to note:
1. By using a `number` for index, you can only access/change indices known at compile time, not runtime (i.e. `myList[0 + 1] = 5` won't work).
2. Not so much an issue right now, but at some point we'll probably want to have a more generalized version of looking up the location to read/assign to for restructured assignment (e.g. `x[0], y = [1, 2]`).

lite-david: Hello from the sets-tuples-dicts group! So, we would require some sort of assign stmt as well to handle setting key-value pairs. myDict[key] = value which looks exactly similar like the list index lookup and assign statement at parse time. I propose we have something like this:
`Stmt<A>:  | {  a?: A, tag: "lookup-assign", name: string, key: Expr<A>, value: Expr<A> }`

At typecheck, the type would be resolved as dict or list and codegen can be done accordingly. How does this sound? @R167

lite-david:
Hello from the sets-tuples-dicts group! So, we would require some sort of assign stmt as well to handle setting key-value pairs. `myDict[key] = value` which looks exactly similar like the list index lookup and assign statement at parse time. I am thinking of something like this:
`Stmt<A>: | { a?: A, tag: "lookup-assign", name: string, key: Expr<A>, value: Expr<A> }`

At typecheck, the type would be resolved as dict or list and codegen can be done accordingly. How does this sound? @R167 @erauyeung

R167:
@lite-david That sounds reasonable so far. For destructuring, we were actually hoping to unify the 3 different assignment types (#16 ids, fields, and eventually index lookup) into one single assign which which takes an `Assignable` expression and shells out to the relevant handler for that particular expression at type checking through tcExpr and again does a similar operation in compiler

For example, the Assignable takes the following form (which is ~similar to how CPython implements this).
```typescript
// Union of all assignable targets
export type Assignable<A> =
  | {  a?: A, tag: "id", name: string }
```

Presently we still maintain the existing "field-assign" and "id-assign" so they can be merged gradually. As long as we can agree on a common interface for how assignment location (e.g. `codeGenIdAssign` assumes the value is should assign is on the top of the stack when it runs, and the same for `codeGenListIndexAssign`, etc.) is determined, everything should work.

lite-david:
Yep, that makes sense.

ast.ts
erauyeung: Yes this one should actually be `Expr<A>`, we simply forgot about it when writing the code 😅

erauyeung:
@ricky8554

ricky8554:
@R167 I think it is a great structure to follow. We will update our structure once we know what the index lookup interface is.

ast.ts
```diff
@@ -37,6 +39,8 @@ export type Expr<A> =
   | {  a?: A, tag: "lookup", obj: Expr<A>, field: string }
   | {  a?: A, tag: "method-call", obj: Expr<A>, method: string, arguments: Array<Expr<A>> }
   | {  a?: A, tag: "construct", name: string }
+  | {  a?: A, tag: "list_expr", contents: [Expr<A>] }
```
jpolitz, line 42: Should that be `Array<Expr<A>>`?

ast.ts
```diff
@@ -37,6 +39,8 @@ export type Expr<A> =
...
+  | {  a?: A, tag: "list_lookup", list: Expr<A>, index: number }
```
jpolitz, line 43: `index` should be an `Expr<A>` I think.

ricky8554: Yes, should be fixed now.

ast.ts
```diff
@@ -37,6 +39,8 @@ export type Expr<A> =
   | {  a?: A, tag: "lookup", obj: Expr<A>, field: string }
   | {  a?: A, tag: "method-call", obj: Expr<A>, method: string, arguments: Array<Expr<A>> }
   | {  a?: A, tag: "construct", name: string }
+  | {  a?: A, tag: "list_expr", contents: Array<Expr<A>> }
```
jpolitz, line 42: Any reason for breaking with the `kebab-case` style in favor of `snake_case`? Sorry I missed that last night.

ast.ts
```diff
@@ -37,6 +39,8 @@ export type Expr<A> =
...
+  | {  a?: A, tag: "list_lookup", list: Expr<A>, index: Expr<A> }
```
jpolitz, line 43: Note that this is going to have to parse exactly the same as this dict bracket lookup form from the dict group. These AST forms should be harmonized to something like `bracket-lookup` and `bracket-assign`, since we can't tell them apart at parse-time.

https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/pull/24/files#diff-08420cd45e72258159c0a8701934f95b0e70129064a4867914f3a3b5782a0e5fR41

@lite-david these groups should agree on these and we can merge what you decide on.

lite-david: @erauyeung @jpolitz  Let's settle on this?  `| {  a?: A, tag: "bracket-lookup", obj:Expr<A>, key:Expr<A> }`  @erauyeung

ricky8554: @lite-david sure, we have updated our ast accordingly. For assignment, we have temporary change the ast to `| {  a?: A, tag: "bracket-assign", obj: Expr<A>, key: Expr<A>, value: Expr<A> }` before the new structure come out.

jpolitz:
Can you retarget this to `main`?

PR #11, opened 2021-02-25
Project proposal for inheritance

Freyr-Wings:
This is our team's proposal for inheritance. Our project is based on this repo: https://github.com/Freyr-Wings/pa1-ucsd-cse231-w21. So our code commits will also be in this repo.


jpolitz:
This is great! I assume this is going to be developed independently of the main project.

One though/request – is there a way on your repo to copy our deploy steps so we can see all the cool things you've done?

Freyr-Wings:
I keep the interface the same as it is in PA3 for testing, and replace all the other config files with the counterpart in this repo. What deploy steps should I add specially for this repo?

PR #15, opened 2021-02-25
Project proposal for fancy calling conventions

danielytan:
This is the project proposal for fancy calling conventions; we include the details in `fancy-calling-conventions.md` and made tiny changes to `ast.ts`, `parser.ts`, and added a single simple test for a function with a default argument in `runner.test.ts`


package.json
```diff
@@ -7,7 +7,7 @@
     "start": "webpack serve --mode=development --open",
     "build-web": "webpack",
     "build-parse": "tsc parsestart.ts --esModuleInterop --moduleResolution node --outDir cli/",
-    "test": "env TS_NODE_COMPILER_OPTIONS='{\"module\": \"commonjs\" }' node --experimental-wasm-bigint node_modules/.bin/mocha -r ts-node/register --reporter mochawesome 'tests/**/*.test.ts'"
+    "test": "env TS_NODE_COMPILER_OPTIONS='{\"module\": \"commonjs\" }' node --experimental-wasm-bigint node_modules/mocha/bin/mocha -r ts-node/register --reporter mochawesome 'tests/**/*.test.ts'"
```
jpolitz, line 10: Why change the test command?

parser.ts
```diff
@@ -347,8 +347,15 @@ export function traverseParameters(c : TreeCursor, s : string) : Array<Parameter
     c.nextSibling(); // Focuses on type itself
     let typ = traverseType(c, s);
     c.parent();
-    c.nextSibling(); // Move on to comma or ")"
-    parameters.push({name, type: typ});
+    c.nextSibling(); // Move on to comma or ")" or "="
+    nextTagName = c.type.name; // NOTE(daniel): copying joe's hack for now
```
jpolitz, line 351: Oh no! Don't copy Joe's hack.

fancy-calling-conventions.md
```diff
@@ -0,0 +1,175 @@
...
+
+### Keyword arguments
+For testing keyword arguments, we will switch the argument positions but ensure that the output of the function stays the same.
+
+### Regular arguments
+To ensure the basic functionality of method calls and function calls are still intact, we will write the exact program without the new calling conventions and make sure the outputs are correct.
+
+## New AST forms:
+
+To account for parameters with default arguments, we would need to include a new pattern for the `Parameter` type in `ast.ts`. The new pattern should include a field for the default value.
```
DDEle, line 145: Assume that the `Type` for callable is something like `Callable[[int, bool], int]` in Python and below in AST
``` typescript
export type Type = ...
    | {tag: "callable", args: Array<Type>, ret: Type}
```
How should it be changed to enable default arguments?

fancy-calling-conventions.md
```diff
@@ -0,0 +1,175 @@
...
+Additionally, we would need to include a field in both call and method call expressions to account for the inclusion of keyword arguments. The `kwargs` field should correspond to an array of arrays of argument names followed by their values.
+```
+export type Expr<A> =
+| { a?: A, tag: "call", name: string, args: Array<Expr<A>>, kwargs: Array<[string, Expr<A>]> }
+| { a?: A, tag: "method-call", obj: Expr<A>, method: string, args: Array<Expr<A>>, kwargs: Array<[string, Expr<A>]> }
+```
+
+## New functions, datatypes, and/or files:
+
+In our global environment, we would need to add another map for functions which would keep track of the parameters names and its corresponding default values. We need to keep track of these fields in order to properly populate the arguments of function calls during compile time.
```
DDEle, line 161: As a member of the closures (as well as first-class functions and lambda expression) group, I would say that it seems very difficult to know the number of arguments of a callable at compile time when involving default arguments and first-class functions. Consider the following example:

``` python
f0: Callable[[], int]

# assuming there are 3 callables of type below:
funcbody: Callable[[int], int] = None
expr1: Callable[[], int] = None
cond1: Callable[[], bool] = None

# Above 3 callables are somehow assigned to non-trivial functions/lambda expressions

def f(x:int = expr1())-> int:
    return funcbody(x)

def g()-> int:
    return 1

# maybe some code here that changes the output of expr1() before and after it.  

if cond1():
    # Although `f0` is a `Callable[[], int]` without optional args which is not of  the identical type of `f`,
    # It is very boring if we decide to not support this kind of conformance of types
    f0 = f  
else:
    f0 = g

f0()
```

fancy-calling-conventions.md
xiexr151e: Default arguments shouldn't change for callables.
For args and kwargs, it can change.
I feel like args and kwargs can be optional... args will just be another `Array<Type>`, and kwargs can be `Map<string, Type>`... Does that make sense?

I envision something like:
```
{tag: "callable", args: Array<Type>, * args: Array<Type>, ** kwargs: Map<string, Type>, ret: Type}
```

package.json
danielytan: Oops, sorry! The path for mocha is slightly different for me so I had to change this to make the tests work (I think it's because I'm on Windows?)

I'll make another commit to revert this!

jpolitz: Hm if it works on *nix and it's necessary for windows, we could leave it. Really I just wanted to know. It works fine on my mac so let's leave the change.

jpolitz (approved):
Looks good.

fancy-calling-conventions.md
DDEle: > Default arguments shouldn't change for callables.

I don't really understand this. For example, the `foo` in  `def foo(int:x = 1) -> int:` (func body omitted) should be of type `Callable[[int], int]`, but it should also be of type (able to be conformed/derived as) `Callable[[], int]`. There must be some changes to deal with that problem.

xiexr151e: For your example, I think `Callable[[int], int]` would be the correct typing. 
Do you foresee problems occuring in the typecheck? That's the only place I think problems could arise (specifically checking the arguments against the predefined parameter types).

fancy-calling-conventions.md
xiexr151e: Thinking about ChocoPy, I think it is probably in our best interests to restrict what gets put inside the default arguments.
I'm thinking of default arguments similar to the init statements in the previous PAs - that is, we should only be able to initialize a variable with a literal value.

@danielytan

fancy-calling-conventions.md
DDEle: Well, if you give the type `Callable[[int], int]` to `def foo(int:x = 1) -> int:`, I guess you assumeed that `Callable[[int], int]` can be uses as an callee that requires **one or less** arguments. In that case, the following would be considered as type correct while it is actually not.

``` python
id: Callable[[int], int] = None

def f(x:int) -> int:
  return x

id = f
id()  # should raise an error during type checking
```

fancy-calling-conventions.md
DDEle: Thanks for your reply. The restriction will probably make life easier. We can come back to the discussion of storing parameters' default values after our group settling down the representation of callables. You are no more welcome to keep an eye and give suggestions on our PR thread if you wish. https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/pull/23#discussion_r584003981

fancy-calling-conventions.md
xiexr151e: Yeah, that would not be correct.
If it were something like this though, it would be correct:
```
def f(x:int = 1) -> int:
    return x
id = f
id()
```
Here, an argument is not supplied to `id`, so the definition of `f` will use the default argument.

xiexr151e: @DDEle 

Hey there, sorry for the late reply. Just want to give you guys some updates on what me, @danielytan and @DavidLiu0304 are doing.
Our work hinges on knowing what the type environment is. Therefore, much of our work for default arguments is done in type check. For default arguments, if the user does not supply arguments but default arguments are present, we insert the default arguments in the call expression using the decorated AST technique.
(We are also using this similar technique to implement keyword arguments.)

Now, how do I think this will affect your work on closures? Suppose we have something like this:
```
def f(x : int, y : int = 2) -> int:
    return x + y
g = f
print(g(1)) # prints 3 
print(g(1, 3)) # prints 4
```
Just like as if the `y` parameter is required, `f` in the above snippet should still be of type `Callable[[int, int], int]`, even though we can give callee less than the number of arguments defined in the function because of default functions. In our type check, we will see a call expression and look up its corresponding type information. In the first call, we see that we need default arguments and thus we edit the AST to insert the default argument. In the second call, since we don't need additional default arguments, nothing changes in the AST.
The advantage to this approach is that it doesn't change the code in `compiler.ts`, because once the default arguments are inserted as needed, the function call itself is no longer anything special.
Let me know if you have any questions. Thanks.

DDEle: @xiexr151e Thanks for updating. I'm still worried that if we can combine the default parameter with the first level functions. Considering this example
``` python
def fd2(x : int, y : int = 2) -> int:
    return x + y

def fd3(x : int, y : int = 3) -> int:
    return x + y

g = None
if (a_nontrivial_expression_of_boolean):
    g = fd2
else:
    g = fd3

print(g(1)) # prints 3 or 4?
```

fancy-calling-conventions.md
xiexr151e: @DDEle 
Sorry for the late reply, but we are trying to do without the restriction.
We have a discussion going on in this thread [here](https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/pull/51#discussion_r587713692). Feel free to take a look.

PR #17, opened 2021-02-25
Project proposal for iterators/for loop

Quincy0v0:
This is our tentative project proposal. Please understand that we may still make some changes before the deadline.


Quincy0v0:
Project proposal for iterators/for loop

jpolitz:
In your final version, please do go ahead and add some AST forms if you have the ones you think you need! It's useful to see the actual changes to the AST across all the projects.

Can you retarget your PR to `main`, as well, so once we approve and merge others will be able to easily incorporate changes?

Quincy0v0:
Hi professor Joe, we don't quite understand what you're referring to by "add some AST forms". Do you mean the changes we make to the AST?

Quincy0v0:
> In your final version, please do go ahead and add some AST forms if you have the ones you think you need! It's useful to see the actual changes to the AST across all the projects.
> 
> Can you retarget your PR to `main`, as well, so once we approve and merge others will be able to easily incorporate changes?

Hi professor Joe, we don't quite understand what you're referring to by "add some AST forms". Do you mean the changes we make to the AST?

ast.ts
```diff
@@ -25,6 +25,9 @@ export type Stmt<A> =
   | {  a?: A, tag: "while", cond: Expr<A>, body: Array<Stmt<A>> }
   | {  a?: A, tag: "pass" }
   | {  a?: A, tag: "field-assign", obj: Expr<A>, field: string, value: Expr<A> }
+  | {  a?: A, tag: "for", name: Expr, index?: Expr, iterable: Expr, body: Array<Stmt<A>> }
```
jpolitz, line 28: Should `name` be a string, not an `expr`? Eventually it might be something like an `Assignable` if we have destructuring in this position. But I don't think you can put, say, a field lookup in this position.

jpolitz (changes requested):
Make sure you respond to my comment about the `name` in `for` loops, then we can merge.

ast.ts
Quincy0v0: I've changed the `name` field into a string.

jpolitz: Oh, and these need to be `Expr<A>` (see failing build below). Sorry I missed that in the first review.

Quincy0v0: Thanks for noticing! I've resolved the conflict with `main` branch and passed the checks.

jpolitz (approved):

PR #19, opened 2021-02-25
Fix required else statement parsing

R167:
The parser seems to implement the PA3 grammar. This requires an else block for all if statements.

Example of program that currently returns unexpected output:

```python
if False:
  print(True)
```

Expected: ""
Got: "True"

This is because we advance the cursor to where we expect the else block to be. However, since we don't do any validation that it's actually progressed to the else block, so we end up processing the "then" block again for the "else" section.


jpolitz (approved):
Thanks, this is better than silent undefined behavior in this case!

PR #20, opened 2021-02-25
Front end Proposal

shanliyu:
Team member: 
Shanli Yu
Ethan Hsu
Yufei Mao


jpolitz:
I see a _lot_ of files added in this commit. It seems like some of these should be possible to include via `npm` and not have in the repo. For example, `CodeMirror` has an `npm` package. I'm also not sure if some of these are built files vs. new source files we need. Thanks!

shanliyu:
Sorry that was a rookie mistake.. Now it has been added to `npm`. Thank you!

index.html
```diff
@@ -3,6 +3,14 @@
...
     <meta charset="utf-8" />
     <script src="https://cdn.jsdelivr.net/gh/AssemblyScript/wabt.js@1.0.20/index.js"></script>
+    <link rel="stylesheet" href="../theme/css/test.css">
+    <link rel="stylesheet" href="../node_modules/codemirror/lib/codemirror.css">
+    <link rel="stylesheet" href="../node_modules/codemirror/theme/neo.css">
+    <script src="webstart.js"></script>
+    <script src="../node_modules/codemirror/lib/codemirror.js"></script>
+    <script src="../node_modules/codemirror/addon/edit/closebrackets.js"></script>
+    <script src="../node_modules/codemirror/mode/python/python.js"></script>
+    <script src="../theme/js/test.js"></script>
```
R167, line 13: I'm somewhat suspicious of including codemirror by directly referencing from node_modules. Was it not possible to include through webstart and making sure css is being loaded through webpack? (Codemirror docs on using a bundler. While for Rollup, some of this should still be reasonably transferrable https://codemirror.net/6/examples/bundle/)

index.html
```diff
@@ -13,7 +21,7 @@
             margin: 0;
             padding: 0;
             height: 100%;
-        }
+        }z
```
R167, line 24: This feels like a typo? (or is there some obscure css i'm missing here?)

package.json
```diff
@@ -28,7 +28,25 @@
...
+    "gulp-bower": "0.0.15",
+    "gulp-eslint": "^6.0.0",
+    "gulp-istanbul": "^1.1.3",
+    "gulp-less": "^4.0.1",
+    "gulp-livereload": "^4.0.2",
+    "gulp-minify": "^3.1.0",
+    "gulp-notify": "^3.2.0",
+    "gulp-plumber": "^1.2.1",
+    "gulp-ruby-sass": "^4.0.0",
+    "gulp-sass": "^4.1.0",
```
R167, line 46: Are all of these really necessary? For example, I just saw [gulp-ruby-sass](https://www.npmjs.com/package/gulp-ruby-sass) is deprecated (2 years ago) in favor of only using gulp-sass. I've never used gulp before, so this is more a question to clarify. In particular, I also wonder how gulp as a stream builder interacts here with webpack as a bundler (e.g. gulp is only being used for theming and is better choice than webpack/works with webpack because X).

R167 (commented):
Couple minor questions about how these stylesheets/additional JS sources are being included.

index.html
shanliyu: Yep this is a typo. Thanks for pointing it out

package.json
shanliyu: I added the packages from another project of mine, at the time there were some features that are not completely migrated to `gulp-sass` so I needed to specify both. For the sake of this project I would not expect such issues. Will remove `gulp-ruby-sass`

index.html
shanliyu: You are right that this is not the best practice of exposing server directory structure. For the sake of using stylesheets and JS I did not expect this to be an issue but you are right this is less than ideal. I can look into the docs you linked but FYI bundling could affect our future implementations for other UI features. Thanks for the info!

jpolitz: Yeah this will actually break the deploy as-is, I think, because we don't ship a full source repo, just the `build/` directory, to the deploy repo. I'm not bothered by direct links like this, but may need to at least copy the files into `build/` as part of `npm run build-web`.

proposals/front-end.md
```diff
@@ -0,0 +1,55 @@
...
+
+11. Multiple editor tabs: <br>
+We are still exploring details regarding whether this is possible to implement. Therefore this is outside of our 10 objectives but we list it here in case we are able to implement in the future. 
+
+
+## How to add tests for features
+Due to the nature of front-end development, while we will definitely attempt to write test cases for our JavaScript, we will probably dedicate more time on QA rather than writing test cases. 
+
+## New AST to Add
+We do not expect to add new AST forms while implementing front-end features.
```
jpolitz, line 42: What about adding location information to the AST?

proposals/front-end.md
```diff
@@ -0,0 +1,55 @@
...
+## How to add tests for features
+Due to the nature of front-end development, while we will definitely attempt to write test cases for our JavaScript, we will probably dedicate more time on QA rather than writing test cases. 
+
+## New AST to Add
+We do not expect to add new AST forms while implementing front-end features. 
+
+## Changes to Existing Functions, Datatypes or Files
+Existing functions: There will be changes made to files such as `webstart.ts` to implement our tooltip design when the cursor is hovering over a printed output. These changes to files will be mainly to HTML attributes. 
+
+Datatypes: An example of datatype changes is that if we are to implement autocomplete for variable and class names, we will be required to store such information inside the global environment during type checking, which can then be utilized in our autocomplete functions.
```
jpolitz, line 47: Yeah, this might affect the REPL interface as well.

jpolitz (approved):
Sorry to make more work for you, but I think we do need a solution that gets all dependencies in the `build/` directory so we can manage the deploy. The test you can run locally is copying `build/` to a completely fresh location, and loading from there.

package.json
jpolitz: Can you say generally what the `gulp` packages are for? I'm not opposed to having several, but I don't recognize the majority of them.

jpolitz:
Oops, I approved accidentally. I meant to just comment.

index.html
R167: @shanliyu Sounds good. I just want to make sure there is a good, clear reason for including an additional build system. My understanding so far is that gulp is ~only being used to process the scss transformations and provide hot reloading, both of which webpack already provides with the sass-loader plugin https://webpack.js.org/loaders/sass-loader/ and HMR https://webpack.js.org/concepts/hot-module-replacement/.

On the point made by @jpolitz, it's not at all clear how to run the build system for these front end changes since they evidently are not being invoked using webpack and `build-web`.

package.json
shanliyu: The most important usage for the gulp packages such as `gulp-sass` and `gulp-minify` are to assist us styling the page since we will be writing sass. Some packages such as `gulp-sourcemaps` can help us while debugging, especially when we are debugging styling with minified css. Others such as `gulp-eslint` are to help us with javascript, while I expect to use them in the future but they are currently not used by front-end

theme/css/test.css
```diff
@@ -0,0 +1,2 @@
+
+/*# sourceMappingURL=test.css.map */
```
R167, line 2: Continuing on the point about not using `build-web`, i'm not sure why these files (which really look like autogenerated build results) are being committed, rather than generated in the build/dev phases.

index.html
R167: If we are going to load codemirror sources directly, we should probably do it using a cdn (https://www.jsdelivr.com/package/npm/codemirror) and declaring it as an external in `webpack.config.js` in much the same way we already do for `wabt`
https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/blob/3ae392fe4f1d13f2b2432e6fab2ddcc96131d48d/index.html#L5-L8
https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/blob/3ae392fe4f1d13f2b2432e6fab2ddcc96131d48d/webpack.config.js#L18-L20

shanliyu: @R167 To answer your question, I personally think integrating sass transformation into build slows down front-end development significantly, especially when more and more changes are being added to the compiler. But using gulp it is independent from the build process which is much faster for us during development

theme/css/test.css
shanliyu: This file is generated by a gulp task. Currently this is empty as our sass file is pretty empty but will be useful in the future

index.html
shanliyu: @jpolitz Can you elaborate a bit more on the deployment process, such as the server structure? I was assuming the hosting server contains everything including `node_modules` as CI/CD builds the site so I included the stylesheets and script tags this way

R167: > I personally think integrating sass transformation into build slows down front-end development significantly

I agree that build slows down dev. That's we we're using webpack-dev-server now. https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/pull/1 Rebuilds take <0.5s. (launch using `npm start` as mentioned in the README).

> I was assuming the hosting server contains everything including `node_modules` as CI/CD builds the site so I included the stylesheets and script tags this way

No, `npm run build-web` only runs `webpack` in production mode. The gh-pages deploy phase then pushes the contents of the `./build` directory to the `gh-pages` branch which is then severed by GitHub pages.

R167: example branch which moves code mirror and sass to being loaded by webpack https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/compare/main...R167:frontend-example

R167: example branch which moves code mirror and sass to being loaded by webpack https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/compare/main...R167:frontend-example

shanliyu: Thanks for the help but this branch does not seem to build
```ERROR in ./webstart.ts 6:0-22
Module not found: Error: Can't resolve './style.scss' in '/Users/shanliyu/School/CSE231/chocopy-wasm-compiler'
resolve './style.scss' in '/Users/shanliyu/School/CSE231/chocopy-wasm-compiler'
  using description file: /Users/shanliyu/School/CSE231/chocopy-wasm-compiler/package.json (relative path: .)
    Field 'browser' doesn't contain a valid alias configuration
    using description file: /Users/shanliyu/School/CSE231/chocopy-wasm-compiler/package.json (relative path: ./style.scss)
      no extension
        Field 'browser' doesn't contain a valid alias configuration
        /Users/shanliyu/School/CSE231/chocopy-wasm-compiler/style.scss doesn't exist
      .ts
        Field 'browser' doesn't contain a valid alias configuration
        /Users/shanliyu/School/CSE231/chocopy-wasm-compiler/style.scss.ts doesn't exist
      .js
        Field 'browser' doesn't contain a valid alias configuration
        /Users/shanliyu/School/CSE231/chocopy-wasm-compiler/style.scss.js doesn't exist
      as directory
        /Users/shanliyu/School/CSE231/chocopy-wasm-compiler/style.scss doesn't exist```
I am exploring the option you provided by using CDN rather than locally for codemirror, and I will still attempt to use gulp to concatenate and minify CSS/JS which will be copied into `/build`.

Comment (R167) [reply index.html:13]: @shanliyu What checkout/build commands did you run to get that to happen? Running from that branch worked fine on my end(?)
```
git remote add R167 git@github.com:R167/chocopy-wasm-compiler.git
git fetch R167
git checkout R167/frontend-example
npm start
```

package.json
jpolitz: Can we leave out the ones that we aren't using yet, just to keep the history of reasons we added things to `package.json` clear? Thanks for going through all this build stuff (typical web annoyances).

jpolitz:
@shanliyu somewhere you asked about how deploy works. We run `npm run build-web` and then copy _just_ the `build/` directory to the `gh-pages` branch, which loads things as static files. So the constraint (which I think is a good one) is that everything that goes live on the deploy needs to be in there.

index.html
shanliyu: So instead of checking to your branch, I copied all of your files and modified my current repo (6 files changed from the link), but when running `npm run build-web` the above error occurs. I tried switching to your branch and it worked without any issues so a bit confused

R167: best guess is you maybe missed a line or didn't restart webpack after npm install? if you want to use any of this (merging probably easiest), feel free.

shanliyu: Really appreciate the help Winston, after merging with the branch it seems that the error is resolved. Not sure why it was occurring before. Thanks again!

package.json
shanliyu: Since I'm a pretty new to webpack after Winston's help now the site should be using webpack instead of gulp, so all the gulp packages are now removed. Learned a lot from Winston's suggestions and branch, kudos to him!

R167: Awesome @shanliyu, glad I could help! note: might still need to cleanup the last couple files:
- theme/gulpfile.js
- theme/css/test.css
- theme/css/test.css.map
- theme/js/test.js

jpolitz:
Cool! From what I know gulp is a fine tool, too, it's just good to agree on one. Thanks for working through this.

jpolitz:
Something odd going on with the merge where it's listing lots of changes. Can you do an explicit `git merge main` on your branch with the up-to-date commits so it doesn't show all the history?

proposals/front-end.md
shanliyu: I think this might be handled by error team based on yesterday's discussion but will check with them first

proposals/front-end.md
shanliyu: I think it will affect the REPL interface but our changes to existing functions in 

> Something odd going on with the merge where it's listing lots of changes. Can you do an explicit `git merge main` on your branch with the up-to-date commits so it doesn't show all the history?

Sorry. Just merged with main. Thank you!

shanliyu:
> Something odd going on with the merge where it's listing lots of changes. Can you do an explicit `git merge main` on your branch with the up-to-date commits so it doesn't show all the history?

@jpolitz Sorry forgot to do that before I pushed, just merged with master. Thank you!

PR #22, opened 2021-02-26
Strings Project Proposal

suhas-goutham:
We have added a project_proposal.md file and made a few changes to the parser to allow parsing of string literals.


jpolitz:
Thanks! Can you retarget your merge at the `main` branch rather than your feature branch? It's easier for us to manage the process that way, and it will make sure you see differences between your code and anything that's updated on our `main`.

suhas-goutham:
We just tried creating a PR targeted at the main branch. But a few test cases are failing in npm tests. That's because we have only made changes in the parser and not in the compiler.

suhas-goutham:
I created a new PR targeted at the main branch. Can you please check if it's correct now ? 
Sorry for any inconvenience.

R167:
@suhas-goutham the new PR looks correct. In the future, you can also always change the base branch of a PR by clicking "Edit" next to the PR Title and changing the "base" dropdown (next to "<author> wants to merge X commits into <base>" https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/changing-the-base-branch-of-a-pull-request
You should be good to either close this one, (or close the other and change this one's base branch).

project_proposal.md
```diff
@@ -0,0 +1,229 @@
+## Strings
```
R167, line 1: Note: You might want to move this file to the "proposals/" directory like it says in the [writeup](https://ucsd-cse231-w21.github.io/proposal/) and rename to something like `proposals/strings.md` since all other proposals will be in that directory.

suhas-goutham:
@R167 Thanks for the suggestion! I will change this one's base branch and close the other PR.

project_proposal.md
suhas-goutham: Done :)

ast.ts
```diff
@@ -49,5 +52,9 @@ export enum BinOp { Plus, Minus, Mul, IDiv, Mod, Eq, Neq, Lte, Gte, Lt, Gt, Is,
 export enum UniOp { Neg, Not };
 
 export type Value =
-    Literal
-  | { tag: "object", name: string, address: number}
+    //Literal
```
jpolitz, line 55: Why not still have `Literal`?

ast.ts
```diff
@@ -41,6 +43,7 @@ export type Expr<A> =
 export type Literal = 
     { tag: "num", value: number }
   | { tag: "bool", value: boolean }
+  | { tag: "string", value: string, length: number}
```
jpolitz, line 46: Can we drop the `length` field of this `Literal` and just use `.length` when needed.

ast.ts
suhas-goutham: Resolved in class. We can't have {tag:"string", value:string} in Value since PyValue takes a number parameter.

ast.ts
suhas-goutham: Dropped length.

suhas-goutham:
I have added a new function called "print_str" in webstart.ts. I think npm test is failing since "print_str" isn't in the main branch's webstart.ts file. So, npm test throws an import error. Any possible fix for this issue ?

ast.ts
```diff
@@ -49,5 +52,9 @@ export enum BinOp { Plus, Minus, Mul, IDiv, Mod, Eq, Neq, Lte, Gte, Lt, Gt, Is,
 export enum UniOp { Neg, Not };
 
 export type Value =
-    Literal
-  | { tag: "object", name: string, address: number}
+    //Literal
+    { tag: "num", value: BigInt }
+  | { tag: "bool", value: boolean }
+  | { tag: "string", value: number}
```
jpolitz, line 58: There's an interesting inconsistency here where `BigInt` is expected to convert its result (even if heap allocated) but strings aren't. Let's use `Literal` again here, and figure out how to return a string. If you need a hack to get started, you can always return strings as `class`-tagged values to start and later do the conversion.

compiler.ts
```diff
@@ -209,6 +211,28 @@ function codeGenExpr(expr : Expr<Type>, env: GlobalEnv) : Array<string> {
       var callName = expr.name;
       if (expr.name === "print" && argTyp === NUM) {
         callName = "print_num";
+      } else if (expr.name === "print" && argTyp === STRING) {
+        callName = "print_str";
```
jpolitz, line 215: I don't think we should run a loop each time we call this function. Can we just pass the address to print and have `print` responsible for finding these characters on the heap and printing out a single string? I don't think we want thousands of calls to `print` for printing a long string.

jpolitz: That would also avoid the need to have these extra `$` variables in the generated code.

You will need to do some research/ask us about how to convert from memory into a JS string from JS, but that is totally doable.

compiler.ts
```diff
@@ -291,10 +315,32 @@ function codeGenExpr(expr : Expr<Type>, env: GlobalEnv) : Array<string> {
   }
 }
 
-function codeGenLiteral(literal : Literal) : Array<string> {
+function allocateStringMemory(string_val:string, env: GlobalEnv) : Array<string>{
+  var newOffset = env.offset;
```
jpolitz, line 319: I'm concerned that this offset won't update dynamically with allocations. In our object allocations, we read the offset from the heap using a load from the first word of our memory array. Don't we need to also do that for strings so we don't overwrite earlier memory?

jpolitz: (For example, if we allocate string memory in a loop, there's no way the compiler can in general know how much memory will be allocated total at runtime, so a static `offset` won't work.)

parser.ts
```diff
@@ -222,7 +230,7 @@ export function traverseStmt(c : TreeCursor, s : string) : Stmt<null> {
       c.nextSibling(); // go to value
       var value = traverseExpr(c, s);
       c.parent();
-
+//a:str="Joe"
```
jpolitz, line 233: Probably a stray comment to remove.

jpolitz (changes requested):
A few things:

- There's code in codeGen but no tests, which worries me, because I'm not sure if the code we're reviewing runs or not
- I don't think the allocation code works for programs that allocate multiple strings

Let's simplify this to _just_ the AST additions with no code other places so we can merge that (and others might find it useful to see it), then we can focus on getting the implementation to work and tests in place.

suhas-goutham:
> A few things:
> 
> * There's code in codeGen but no tests, which worries me, because I'm not sure if the code we're reviewing runs or not
> * I don't think the allocation code works for programs that allocate multiple strings
> 
> Let's simplify this to _just_ the AST additions with no code other places so we can merge that (and others might find it useful to see it), then we can focus on getting the implementation to work and tests in place.

Got it. I assumed since we were making all the changes on this PR, so we could start pushing changes made in implementation as well on this PR. Now, I realized that we would probably have more PRs where we would push other changes as well. Will push only the appropriate changes from next time.

ast.ts
suhas-goutham: Done. I have changed it to Literal.

compiler.ts
suhas-goutham: Got it. So, is there a way to access the js memory in webstart.ts ?(Can you provide any links to some documentation if you know any) 
You're right, that way I can just send the start address by calling print_str once and it prints out the string.
Thanks!

jpolitz:
Thanks! I took the liberty of removing `string-index` and allowing the dict group's `bracket-lookup` to take its place. The type of the value before the `[` can be used to determine which kind of lookup is happening.

R167:
Note: adding the string type won't be handled currently in 3 places where we stringify types:

https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/blob/cdcc9d759254d512090d9e025cc3721bc0f286bd/tests/import-object.test.ts#L4-L15

https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/blob/cdcc9d759254d512090d9e025cc3721bc0f286bd/utils.ts#L3-L14

https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/blob/cdcc9d759254d512090d9e025cc3721bc0f286bd/webstart.ts#L12-L23

Something to be aware of for handling a new type. Initially found this issue in #23 closure proposal. (found with `@typescript-eslint/switch-exhaustiveness-check` from ESLint rules in #3).

suhas-goutham:
> Thanks! I took the liberty of removing `string-index` and allowing the dict group's `bracket-lookup` to take its place. The type of the value before the `[` can be used to determine which kind of lookup is happening.

Thanks! We will also use the 'bracket-lookup' to perform string indexing.

suhas-goutham:
> Note: adding the string type won't be handled currently in 3 places where we stringify types:
> 
> https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/blob/cdcc9d759254d512090d9e025cc3721bc0f286bd/tests/import-object.test.ts#L4-L15
> 
> https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/blob/cdcc9d759254d512090d9e025cc3721bc0f286bd/utils.ts#L3-L14
> 
> https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/blob/cdcc9d759254d512090d9e025cc3721bc0f286bd/webstart.ts#L12-L23
> 
> Something to be aware of for handling a new type. Initially found this issue in #23 closure proposal. (found with `@typescript-eslint/switch-exhaustiveness-check` from ESLint rules in #3).

Yes actually we need to add a new case "string" in the three places you have mentioned. However, in the ast we have:-

```
export type Literal = 
    { tag: "num", value: BigInt }
  | { tag: "bool", value: boolean }
  | { tag: "string", value: string}
  | { tag: "none" }
```
Since, PyValue takes a parameter "result:number", it would be difficult for us to return a "string" to "result". Hence, Joe had mentioned that we can use a hack to return a string class object to PyValue. So, we will work on figuring out how to do that.

suhas-goutham:
@jpolitz It still says "1 change requested". I think I have made all the changes you have asked for. Is there anything else I should be doing before you merge this branch into the "main" branch ?

compiler.ts
suhas-goutham: Thanks for this suggestion. I will look into how we are doing object allocations and do the same for strings.

jpolitz (approved):

suhas-goutham:
Thanks for merging!

PR #23, opened 2021-02-26
Closures proposal

YaosenLin:


jpolitz:
Go ahead and add the small change you have in mind to this PR!

DDEle:
> Go ahead and add the small change you have in mind to this PR!

We appended some code snippets in the *New functions, datatypes, and/or files added to the codebase* part etc. of the proposal.

eair.ts
```diff
@@ -0,0 +1,77 @@
...
+}
+
+export type Closure<A> = { 
+  a?: A, 
+  name: string, 
+  fields: Array<VarInit<A>>, 
+  apply: FunDef<A> 
+}
+
+export type Stmt<A> =
```
jpolitz, line 41:
```
type Assign<A> = { a?: A, tag: "assign", name: string, value: Expr<A> }
type StmtNoDefs<A> = Assign<A> | Return<A> | Expr<A> | ... leave out function def ...
```

ast.ts
```diff
@@ -15,39 +20,57 @@ export type Class<A> = { a?: A, name: string, fields: Array<VarInit<A>>, methods
...
 
 export type Stmt<A> =
-  | {  a?: A, tag: "assign", name: string, value: Expr<A> }
-  | {  a?: A, tag: "return", value: Expr<A> }
-  | {  a?: A, tag: "expr", expr: Expr<A> }
-  | {  a?: A, tag: "if", cond: Expr<A>, thn: Array<Stmt<A>>, els: Array<Stmt<A>> }
-  | {  a?: A, tag: "while", cond: Expr<A>, body: Array<Stmt<A>> }
-  | {  a?: A, tag: "pass" }
-  | {  a?: A, tag: "field-assign", obj: Expr<A>, field: string, value: Expr<A> }
+  | { a?: A, tag: "assign", name: string, value: Expr<A> }
```
jpolitz, line 42: Let's not have big whitespace changes in this PR.

ast.ts
```diff
@@ -15,39 +20,57 @@ export type Class<A> = { a?: A, name: string, fields: Array<VarInit<A>>, methods
...
+  name: string, 
+  parameters: Array<Parameter<A>>, 
+  ret: Type, 
+  decls: Array<Scope<A>>,
+  inits: Array<VarInit<A>>, 
+  funs: Array<FunDef<A>>
+  body: Array<Stmt<A>>
+}
+
+export type Closure<A> = {
```
jpolitz, line 34: Agree with adding this here, thanks for moving it.

ast.ts
```diff
@@ -15,39 +20,57 @@ export type Class<A> = { a?: A, name: string, fields: Array<VarInit<A>>, methods
...
-export enum BinOp { Plus, Minus, Mul, IDiv, Mod, Eq, Neq, Lte, Gte, Lt, Gt, Is, And, Or};
+export enum BinOp { Plus, Minus, Mul, IDiv, Mod, Eq, Neq, Lte, Gte, Lt, Gt, Is, And, Or };
 
 export enum UniOp { Neg, Not };
 
 export type Value =
-    Literal
+  | Literal
   | { tag: "object", name: string, address: number}
+  | { tag: "callable", name: string, address: number}
```
jpolitz, line 76: What's the plan for the name of a `lambda`-created function?

proposals/closures.md
```diff
@@ -0,0 +1,256 @@
...
+	  // escape analysis here
+  }
+  ```
+
+## Changes to existing functions, datatypes, and/or files in the codebase
+We need to change the `parser.ts` to support recursively parsing nested fucntions;
+`tc.ts` for the new `callable` type and so on; `compiler.ts` to support the new features added.
+
+## Value representation and memory layout for any new runtime values
+The memory layout of closures and references should be as similar as that of class (reference wrapper and closure implemented as something like non-inheritable class). As for non-escaping functions using as values, they will be an i32 pointer pointing to the corresponding functions in the table.
```
jpolitz, line 253: This will require a little bit of careful design, but not too much I hope. We can use the extra-argument infrastructure for method calls to pass an extra reference into the function that holds all the nonlocal variables.

The main thing is that we don't have vtable support yet, and waiting for the inheritance group isn't a good idea. So maybe just having the first word of the closure as the function pointer would work well. That would also keep their representation working the same as function values that _don't_ close over any nonlocal variables, so the indirect_call's index is always found by loading the first word of memory referenced by the function value (whether closure or regular function).

That sounds like a good place to start.

jpolitz (changes requested):
Just requesting that the whitespace changes _not_ be included so that we don't have major conflicts with other AST edits.

ast.ts
YaosenLin: can you release the formatter? So that we can format before PR. Thanks.

ast.ts
YaosenLin: the lambda created function is compiled to wasm function named `$lambda_n` where `n` is a global counter in the environment. A callable value is created when a function or a lambda is compiled, where its `name` is the function name, and its `address` stores the table index. This callable value is assigned to variable of the function name for ordinary function, and assigned to the callable variable for lambda expression (e.g. `f = lambda x : 0`)

proposals/closures.md
YaosenLin: Great. Regular callable variable is a pointer pointing to the index. The closure object's first field contains the index. This makes function call consistent. For closure object, the rest of the fields contain extra arguments, more stuffs needs to be pushed to stack machine before `call_indirect`.

ast.ts
jpolitz: The correct order is to format **after** these PRs merge. I don't want a bunch of whitespace changes in everyone's PRs, I want everyone to re-pull correctly-formatted code. This changeset should not have whitespace changes.

The setup for the formatter is in https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/pull/3

DDEle:
> Just requesting that the whitespace changes _not_ be included so that we don't have major conflicts with other AST edits.

Finished in https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/pull/23/commits/315ef08984dd9431554ce1888da43bde0c56a7d7

proposals/closures.md
DDEle: > We can use the extra-argument infrastructure for method calls to pass an extra reference into the function that holds all the nonlocal variables.

@jpolitz Did you mean that every compiled wasm function should have exactly one extra parameter (except potential things needed by the fancy-calling group), which is a pointer 4 bytes greater than the function value (or the same as the function value to harness recursion)? It is similar to the `self` arguments in the `apply` method of [closure class implementation](https://ucsd-cse231-w21.github.io/closures/#from-nested-functions-to-closures).
Another operation is to load nonlocal variables one by one outside, before the indirect function call, and passes all of them as separated function arguments, as said by Yaosen in the last sentence.

The first interpretation/implementation is base on the feature that nested functions/closures can be used as `callable` variables, stripping the information of its extra arguments, which makes it impossible to apply different numbers of extra arguments to different functions. Consider the following example: (modified from [nested functions lecture](https://ucsd-cse231-w21.github.io/nested-functions/))
``` python
def f(x : int) -> int:
  def g(y : int) -> int:
    return x + y

  def h(y : int) -> int:
    return 1 + y

  var_callable: Callable[[int], int] = None

  if True:  # suppose in a nontrivial example, the condition is too complex to predict at compile time
    var_callable = g  # the heap memory containing func table index of f_g
  else:
    var_callable = h  # the heap memory containing func table index of f_h, maybe followed by the reference (*if x's value is changed after this line but before calling*) of x

  # now we don't know how many/if necessary separate extra arguments we need to add to var_callable at compile time
  return var_callable(10)

f(5)
```

jpolitz: Right – I don't think this can work with arbitrary numbers of extra arguments for escaping functions with nonlocals. It has to be a single heap location, because of this kind of example.

jpolitz (approved):

jpolitz (approved):

R167:
Note: The `callable` type is currently unhandled in 3 switch statements in the code base:

https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/blob/cdcc9d759254d512090d9e025cc3721bc0f286bd/tests/import-object.test.ts#L4-L15

https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/blob/cdcc9d759254d512090d9e025cc3721bc0f286bd/utils.ts#L3-L14

https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/blob/cdcc9d759254d512090d9e025cc3721bc0f286bd/webstart.ts#L12-L23

We should make sure to add callable to these cases, or at least throw an exception if we try to stringify an unsupported type. It may also be worth merging some of these together (esp. the `utils.test.ts` and `webstart.ts` stringify). Issue caught by proposed ESLint changes in #3 (non-exhaustive switch statement, `@typescript-eslint/switch-exhaustiveness-check`)

DDEle:
> Note: The `callable` type is currently unhandled in 3 switch statements in the code base:
> 
> https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/blob/cdcc9d759254d512090d9e025cc3721bc0f286bd/tests/import-object.test.ts#L4-L15
> 
> https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/blob/cdcc9d759254d512090d9e025cc3721bc0f286bd/utils.ts#L3-L14
> 
> https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/blob/cdcc9d759254d512090d9e025cc3721bc0f286bd/webstart.ts#L12-L23
> 
> We should make sure to add callable to these cases, or at least throw an exception if we try to stringify an unsupported type. It may also be worth merging some of these together (esp. the `utils.test.ts` and `webstart.ts` stringify). Issue caught by proposed ESLint changes in #3 (non-exhaustive switch statement, `@typescript-eslint/switch-exhaustiveness-check`)

Sorry for the problem. I guess that we can throw an exception now as we haven't actually implemented any callable-thing yet. We will replace the exception with something meaningful later.

PR #24, opened 2021-02-26
Proposal and ast changes for implementation of dicts 

lite-david:
This PR contains a few changes to ast.ts to support dicts. It also contains our proposal in proposals/sets-tuples-dicts.md
Team members: Edwin, Dheeraj, Hema, Sudharshan

### DIFF
--- a/ast.ts
+++ b/ast.ts
@@ -39,7 +39,7 @@ export type Expr<A> =
   | {  a?: A, tag: "construct", name: string }
   | {  a?: A, tag: "comprehension", expr: Expr<A>, field: string, iter: Expr<A>, cond?: Expr<A> }
   | {  a?: A, tag: "block", block: Array<Stmt<A>>, expr: Expr<A> }
-  | {  a?: A, tag: "dict", value: Map<Expr<A>, Expr<A>> }
+  | {  a?: A, tag: "dict", entries: Array<[Expr<A>, Expr<A>]> }
   | {  a?: A, tag: "bracket-lookup", obj:Expr<A>, key:Expr<A> }
 
 export type Literal = 
--- a/ast.ts
+++ b/ast.ts
@@ -37,11 +37,10 @@ export type Expr<A> =
   | {  a?: A, tag: "lookup", obj: Expr<A>, field: string }
   | {  a?: A, tag: "method-call", obj: Expr<A>, method: string, arguments: Array<Expr<A>> }
   | {  a?: A, tag: "construct", name: string }
-<<<<<<< HEAD
   | {  a?: A, tag: "comprehension", expr: Expr<A>, field: string, iter: Expr<A>, cond?: Expr<A> }
   | {  a?: A, tag: "block", block: Array<Stmt<A>>, expr: Expr<A> }
-=======
->>>>>>> 6d24f62f2fcbcfee9ffbfb7dfcdb711c911de92f
+  | {  a?: A, tag: "dict", value: Map<Expr<A>, Expr<A>> }
+  | {  a?: A, tag: "bracket-lookup", obj:Expr<A>, key:Expr<A> }
 
 export type Literal = 
     { tag: "num", value: BigInt }


ast.ts
```diff
@@ -25,6 +25,7 @@ export type Stmt<A> =
   | {  a?: A, tag: "while", cond: Expr<A>, body: Array<Stmt<A>> }
   | {  a?: A, tag: "pass" }
   | {  a?: A, tag: "field-assign", obj: Expr<A>, field: string, value: Expr<A> }
+  | {  a?: A, tag: "lookup-assign", name: string, key: Expr<A>, value: Expr<A> }
```
R167, line 28: As discussed in #9 here's a quick link to an idea of the interface we've proposed for fully unifying assignment. Note: we're still using the old `id-assign` and `field-assign` parsing for this initial PR. https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/blob/11cb027e707adaa4c9667bbaf0bf5d8f740e9040/proposals/destructuring.md#why-this-form-of-unified-assignable

lite-david: Thanks! I removed the assign statement from the ast for now. We will wait for your changes to get in, and then use the interface that you've proposed.

R167: Awesome! I'll make sure to tag in once we have something in a subsequent PR.

ast.ts
```diff
@@ -37,6 +37,8 @@ export type Expr<A> =
   | {  a?: A, tag: "lookup", obj: Expr<A>, field: string }
   | {  a?: A, tag: "method-call", obj: Expr<A>, method: string, arguments: Array<Expr<A>> }
   | {  a?: A, tag: "construct", name: string }
+  | {  a?: A, tag: "dict", value: Map<Expr<A>, Expr<A>> }
+  | {  a?: A, tag: "key-lookup", name:string, key:Expr<A> }
```
jpolitz, line 41: Should `name` be an expression? What about `d[k][k2]`?

ast.ts
```diff
@@ -37,6 +37,8 @@ export type Expr<A> =
   | {  a?: A, tag: "lookup", obj: Expr<A>, field: string }
   | {  a?: A, tag: "method-call", obj: Expr<A>, method: string, arguments: Array<Expr<A>> }
   | {  a?: A, tag: "construct", name: string }
+  | {  a?: A, tag: "dict", value: Map<Expr<A>, Expr<A>> }
```
jpolitz, line 40: Does Python have constraints on which order we evaluate key and value expressions? How might it be difficult to maintain this with a `Map`?

lite-david: The latest versions of Python evaluate the key and value pairs in order they appear for dicts. In Js too a `Map` remember the original insertion order of the keys: [JS maps](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map). While parsing we will update the map in order of appearance of the keys in the source. So I think it should be fine. Do you think otherwise?

lite-david:
@jpolitz I pulled in all the changes that went into main to my branch and so it shows a lot of files changed, but the main change is in ast.ts only ie adding expr for dict and bracket-lookup.

ast.ts
jpolitz: Oh interesting! It still is odd to me because I don't think there are compelling reasons to do a `.get` on this map.

jpolitz:
@lite-david it shouldn't show all of those changes... git should know there's a merge from `main` and only show the relevant stuff. How did you move things over?

Pulling changes from https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler `main` should work, or pulling those changes into your local `main` branch and then running `git merge main` while on your feature branch should also work.

I'd rather merge something where the history is clear.

lite-david:
@jpolitz Cleaned it up

ast.ts
lite-david: @jpolitz Yes you're right, I don't see where we'd do a .get so perhaps a map for this is an overkill for now. Do you think it would be better to change it to `keys:Array<Expr<A>>, values:Array<Expr<A>>`, or `value:Array<[Expr<A>, Expr<A>]>` ?

ast.ts
lite-david: Fixed this and it's now bracket-lookup like discussed [here](https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/pull/9#discussion_r583870506)

ast.ts
jpolitz: Probably the last one – easiest to keep them together to evaluate them together, and can do a destructuring assignment inside a map/filter/fold, etc.

jpolitz: Maybe call it `entries`

lite-david: Done. I replaced the map and renamed the field to entries in the latest commit.

jpolitz (approved):
LGTM

proposals/sets-tuples-dicts.md
```diff
@@ -0,0 +1,146 @@
+# Project Proposal: Dictionaries
+
+## 1. 10 Representative example programs related to our feature
+
+By integrating the support from strings and lists, we seek to support dictionary snippets such as: 
+
+
+**Example 1**:
```
jpolitz, line 8: There need to be some more sophisticated examples here. Not putting this as a blocker for a merge, but:

- How will type annotations look for dicts?
- All of these examples use constants in key/value position on literals, lookup, and assignment. How about some with interesting expressions there to make sure type-checking will work?

lite-david: @jpolitz okay yes, we'll update the examples in the proposal. 
Follow up question regarding type in dicts: Does this snippet need to throw an error in our compiler? In regular python in doesn't because after x = d[4] the type of x is int and not bool. But ideally in chocopy this should throw a type error.  
```
x:bool = True
d = {}
d[4] = 2
x = d[2+2]
```

zilu0907:
Hi, I'm from the front-end group, have you guys implemented printing dicts? That may have some overlap with our group.

lite-david:
@zilu0907 Hi! We haven't yet. How do you want to handle the overlap?

zilu0907:
Here's what we implemented for object. 
image
We mainly look into PyValue, and then add html elements to display the members in the object in webstart.js.
We can do similar thing for dicts.
What do you think?

PR #27, opened 2021-02-26
Type Inference and Generics Proposal 

Jake-R-Sutton:


jpolitz:
I don't know that we'll have inheritance in place to do subtyping. I think it's enough to focus on local inference for now.

Can you make some small change to the interfaces/APIs that you think you need? In our meeting in class we talked about adding `infer` which happens _before_ type-checking and adds in the inferred types for statements that don't have them. This will probably also require some loosening of parser restrictions, would be good to see a bit of how you see that fitting in.

proposals/type_inference.md
```diff
@@ -0,0 +1,163 @@
...
+	- Ensure that users can declare variable as a super type to override the type that would have been inferred successfully
+	- Ensure generic types exist and their usage is valid 
+- Check that programs that use type inference produce correct outputs
+- Check that reasonable errors are produced when more type information is required 
+
+*We may also consider exploring property-based testing if time permits. 
+
+## Ast Forms
+
+- For Local Type Inference: No changes should be needed
```
jpolitz, line 129: Is this true? Do you know if you need to change anything about the `VarInit` sequences in the prelude of function bodies?

nliittsc: I think this was written under the assumption that we may not necessarily start from this particular compiler implementation. It seems like you want us to use this compiler, so should we adjust our proposal under this assumption?

nliittsc:
I think it would be helpful if we try to pin down a few definitions in order to make sense of exactly what the scope of the project is/should be (and also to ease communication). @jpolitz 's input on this would be helpful, since we should all get on the same page. Here are the points I think are confusing:

* One thing that is confusing me is the fact that classes in Python are types, which seems to imply that inheritance is *directly* related to sub-typing. Where exactly they are orthogonal is not clear to me.
* Because of the above, when discussing local type inference, we should probably consider what it means to "infer the type" of a class object when doing an assignment. E.g., It *appears* that we can use inheritance to get most of the way there, but I understand that this is actually non-trivial in general. So when @jpolitz says "Lets focus on local inference for now", do you mean that we do *not* infer the types of variables `x` when `x` has been assigned a class? E.g., should `x = Counter()` throw an error for now?

jpolitz:
> One thing that is confusing me is the fact that classes in Python are types, which seems to imply that inheritance is directly related to sub-typing. Where exactly they are orthogonal is not clear to me.

I'm not sure anyone said they are. I think on the call I said that _generics_ are (mostly) orthogonal to _inheritance and subtyping_. While they interact (e.g. type bounds on generics, subclassing generic classes), they are separate features.

> local inference

By “local” I mean that inference should start by assuming that all function parameters have annotations, and inference should proceed only on the bodies of functions.

The reason global inference is difficult in the presence of classes (not even talking subtyping) is, for example, this program:

```
class C(object):
  x : int = 0
class D(object):
  x : int = 0
fun f(o : ___) -> int:
  return o.x
```

What is the type that should be inferred for `o`? I'm not sure.

Since any global inference _also_ requires local inference, I think it's pragmatic to start within function bodies and then scale up.

jpolitz:
(I see you have the example I posted in your proposal, so now I'm even more confused about why starting with local inference is controversial.)

Here's a program to consider that is challenging with subtyping and _local_ inference, and is idiomatic Python.

```
class C(object):
  x : int = 0
class A(C):
  z : int = 0
class D(A):
  y : int = 0
class E(A):
  y : int = 0

def f(n : int):
  if n > 0:
    o = D()
  else:
    o = E()
  return o.x / return o.z / return o.y
```

Consider what types we might assign to `o` depending on which return statement is chosen.

Jake-R-Sutton:
Sorry for the delay in following up @jpolitz! 
- To amend our proposal about AST changes, I believe to support our first 4 Programs here we'll need to change the ```VarInit``` in ```ast.ts``` to allow expressions instead of just ```Literal```s. Then we'll also need to infer those types before type checking as you mentioned in the other comment.
- We also plan to move forward with local inference, as you suggest. 
- A thought that we've discussed in our group chat is to move toward implementing generic types as sets of methods and fields that are used on a variable within a function or method (whose signatures are known), and allowing type checking to ensure those contracts are met by parameters to this function. I believe this could support your concluding example.

jpolitz:
- I think it's more than allowing any expression in `VarInit`, right? Also needs to drop the annotation position. I think really it's that programs `assign` statements that don't have a corresponding `VarInit` are allowed?
- 👍🏻
- That doesn't sound like generics to me, but the goal sounds reasonable! Generics typically mean that there is some notion of a _declared type variable_ that can be _explicitly instantiated_. The inference you're describing is something different, so let's come up with a good word for it (maybe type constraints or something).

nliittsc:
We are actually in a call right now trying to work on some of this stuff.
* We just realized what you meant here. We're trying to settle on a design decision of treating everything as essentially an assignment with optional type annotations, or treating "declarations" and "assignments" differently. We realized that since class objects require some pointers and heap allocations, that we have to think about what allowing for "multiple declarations" means.

E.g., consider this program:
```
x : int = 5
x : int = 5  # Throws an error in chocopy, OK in python3
```

Since we are doing type inference, we can allow a declaration to consist of just `x = 5` or even `x = 2 + 3`. However, we are trying to settle on this behavior:
```
x = 2 + 3   # counts a declaration
x : int = 5  # should this throw an error?
x = True   # Should this throw an error?
x : Bool = 5  # Should this throw an error?
```
IMO, the 2nd line above should be OK, but the 3rd and 4th lines above would throw a type error *if these three occur in the same program*. This is following the convention that typed variables stay the same type during a program's lifetime. A declaration is just the 'first assignment', which allow *optional* type annotations, but those type annotations are *concrete* in the sense that they *will be used in type-checking* which is unlike python3 behavior. Also, there's something to think about with class variables, which are really just pointers to memory locations:
``` 
c : Counter = None   # Is this allocated a pointer?
c : Counter = None   # Should this have a new pointer?
```
At the moment, I think following Python3 and allowing every assignment to also work as a declaration makes sense, and at the moment it shouldn't be *that* hard to drag around an environment which checks if an assignment 'should be a declaration'.

* 👍 
* Honestly, we might have to drop generics. Type inference might be more than enough for now.

Any more thoughts, @jpolitz ?

PR #28, opened 2021-02-26
Run tests for deploy

R167:
In lecture, stated tests should pass for deploying, however this currently is not the case. Additionally, attribute the commits to the `github-actions` bot https://github.com/peaceiris/actions-gh-pages#%EF%B8%8F-set-git-username-and-email.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -22,6 +22,9 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
+      - name: Run tests
+        run: npm test
+
       - name: Build app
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -22,6 +22,9 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
+      - name: Run tests
+        run: npm test
```
jpolitz, line 26: Oh right now that `ci` is used it doesn't run tests.

jpolitz (approved):

PR #29, opened 2021-02-26
Memory Management Proposal

student-of:
Memory Management Proposal.

Includes some changes important to the project (made in new files).


tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "outDir": "./dist/",
     "noImplicitAny": true,
-    "target": "es2015",
+    "target": "es2020",
```
jpolitz, line 5: Can you say a little about why you need this?

jpolitz (approved):
Looks good (mostly new files). Could be nice to not require ES2020, but leaving the update in for now.

tsconfig.json
student-of: Just for `BigInt` literals. We can probably do without because JavaScript integers can represent 32bit integers and we locked ChocoPy values to 32 bits.

PR #30, opened 2021-02-26
Numpy initial submission

bensonlyu:
- added numpy.md under proposals/
- made minor changes to ast.ts and parser.ts for import statements and import class tracking
- created a folder np/ that contains relevant files/examples for numpy implementation


ast.ts
```diff
@@ -44,7 +45,7 @@ export type Literal =
   | { tag: "none" }
 
 // TODO: should we split up arithmetic ops from bool ops?
-export enum BinOp { Plus, Minus, Mul, IDiv, Mod, Eq, Neq, Lte, Gte, Lt, Gt, Is, And, Or};
+export enum BinOp { Plus, Minus, Mul, IDiv, Mod, Eq, Neq, Lte, Gte, Lt, Gt, Is, And, Or, Dot};
```
jpolitz, line 48: I'm confused by this – how does `Dot` differ from field lookup? Is it syntactically any different?

np/numpy.js
```diff
@@ -0,0 +1 @@
...packages/numpy/core/tests/test_longdouble.py",start:13851517,end:13857831,audio:0},{filename:"/lib/python3.8/site-packages/numpy/core/tests/test_scalarinherit.py",start:13857831,end:13859628,audio:0},{filename:"/lib/python3.8/site-packages/numpy/core/tests/_locales.py",start:13859628,end:13861934,audio:0},{filename:"/lib/python3.8/site-packages/numpy/core/tests/test_umath.py",start:13861934,end:13968741,audio:0},{filename:"/lib/python3.8/site-packages/numpy/core/tests/test_memmap.py",start:13968741,end:13976057,audio:0},{filename:"/lib/python3.8/site-packages/numpy/core/tests/test_regression.py",start:13976057,end:14061125,audio:0},{filename:"/lib/python3.8/site-packages/numpy/core/tests/test_umath_complex.py",start:14061125,end:14080448,audio:0},{filename:"/lib/python3.8/site-packages/numpy/core/tests/test_scalar_ctors.py",start:14080448,end:14082742,audio:0},{filename:"/lib/python3.8/site-packages/numpy/core/tests/data/astype_copy.pkl",start:14082742,end:14083458,audio:0},{filename:"/lib/python3.8/site-packages/numpy/core/tests/data/recarray_from_file.fits",start:14083458,end:14092098,audio:0}],remote_package_size:7482064,package_uuid:"bc18a0cb-fd54-4271-bcf7-bf29a53ba511"})})();
```
jpolitz, line 1: Is this available as an `npm` module? I'd rather not commit the blob of source to the repo if it's installable in another way.

proposals/numpy.md
```diff
@@ -0,0 +1,162 @@
+## Introduction
+
+This project implements a subset of Numpy with focus on the most popular scalar and matrix operations. Two alternative solutions will be explored:
+
+- Run commands in WASM via [Pyodide](https://github.com/iodide-project/pyodide), which interfaces compiled Numpy packages in WASM.
```
jpolitz, line 5: I'm also curious if we could install Pyodide separately, rather than checking in the whole thing (it's a big dependency!). Is it possible to just use the numpy part of Pyodide without importing all of it?

jpolitz (commented):
Need some feedback/decisions on what code we will be including in the repo before merging this. Great ideas for getting started!

ast.ts
R167: Reading the proposal, I believe this is being used to signify the matmul `@` operator in python. https://docs.python.org/3/library/operator.html#operator.matmul. The name is a little confusing though at first, esp since it's a bit more obscure syntax for vanilla python

bensonlyu: Yes. `Dot` is shortened for '"dot product of matrixes" or Numpy method `a.dot(b)` or `@` operator, which I thought is a very cool operator. Do you need me to change the enum to something like `Matmul`, `At`, etc.?

np/numpy.js
bensonlyu: This is not available in `npm`. Will remove it for future pull requests.

proposals/numpy.md
bensonlyu: I didn't find an existing approach to install the Numpy part of Pyodide separately. Will look around further. Thanks for all these suggestions professor!

ast.ts
jpolitz: Ooooohhh. Sorry I read the code and the proposal separately and didn't connect the.... dots.

Yeah let's call it MatMul, since that's the name of the dunder method. Thanks, sorry for the think-o on my part.

jpolitz:
Can you retarget this to `main`?

And to clarify/confirm – we have to have these big JS files in our repo? If we do, we should include a corresponding `LICENSE` file in the `np` subdirectory.

Do we need the `np` index file? That looks like a test page. And what's the empty `.data` file?

I think an overall description of what these files are for could help. If we really have to load all of pyodide it seems like this isn't a great approach.

bensonlyu:
Yes, I will retarget it to `main` once confirming the import statement syntax and parsing with another builtin team.

These JS files are not necessary to be in the repo now. But if they do later, I will add a `LICENSE`.

`index.html` in `np` is not needed, which was just for testing.

`numpy.data` is a compiled WASM file of the Numpy package.

Will remove `np/` and all files within.

PR #33, opened 2021-02-27
Fix caching action so cypress binary is cached (consequence of Testing PR #5)

R167:
In the Testing PR #5, Cypress was added as a dependency. This downloads and installs the cypress binary if it doesn't exist, or uses the cached version if it does. Add the cypress cache directory to speed up tests during the install phase (needing to reinstall cypress each time added ~30 seconds).

Docs on cypress cache location: https://docs.cypress.io/guides/getting-started/installing-cypress.html#Binary-cache
Example action run verifying that cypress correctly uses the cache to restore: https://github.com/R167/chocopy-wasm-compiler/runs/1992137958?check_suite_focus=true#step:4:11

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -14,7 +14,9 @@ jobs:
       # Make use of npm install cache
       - uses: actions/cache@v2
         with:
-          path: ~/.npm
+          path: |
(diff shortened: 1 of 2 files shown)


jpolitz (approved):
LGTM

PR #34, opened 2021-02-27
BigInt type bug fix

lsteffkojr:
In typescript, the correct type declaration for BigInt values is actually 'bigint'. Interestingly, using 'BigInt' instead of 'bigint' seems to pass test cases and work correctly when run locally, but fails on the live project site.

### DIFF
--- a/ast.ts
+++ b/ast.ts
@@ -44,7 +44,7 @@ export type Expr<A> =
   | {  a?: A, tag: "block", block: Array<Stmt<A>>, expr: Expr<A> }
 
 export type Literal = 
-    { tag: "num", value: BigInt }
+    { tag: "num", value: bigint }
   | { tag: "bool", value: boolean }
   | { tag: "none" }
 


jpolitz:
It might be a `tsconfig` issue. Can you link to a failing build/explain how to reproduce this?

lsteffkojr:
It turns out the live project site was failing because I didn't realize my web browser was out of date. However, this is still an issue. There is no current build that is affected, but if we (or another group) add code that performs any arithmetic operation on the literal value stored in the AST, it will fail to compile, as in the attached example. This will matter because our project will add a function that will use arithmetic operations to separate the BigInt value in the AST into 30-bit pieces so they can be stored in WASM.
Screen Shot 2021-02-27 at 13 52 33
Screen Shot 2021-02-27 at 13 42 41

jpolitz (approved):
Thanks for explaining!

PR #36, opened 2021-02-28
Parsing for Dicts

sudharshan-chakra:
Parser.ts changes - parsing dictionary expressions and bracket-lookup.
@lite-david @dheeraj7596 @hemathota


lite-david:
@ricky8554 @erauyeung as part of this PR we've already added parsing for bracket-lookup Exprs. Pointing it out to avoid double efforts.

jpolitz (dismissed):
I would rather not merge untested implementation code. Can we get some tests and some “not yet implemented” exceptions in the relevant places?

jpolitz (changes requested):
I'd rather not merge untested implementation code. Can we get some parse tests and some “not yet implemented” exceptions in the relevant places?

jpolitz:
**Thanks** for submitting an incremental pull request!

lite-david:
@jpolitz I added parser tests in tests/parser.test.ts. 
>  “not yet implemented” exceptions in the relevant places?

In typecheck and codegen the unhandled cases already default to `unhandledTag` which throws an exception. Should we be doing something more there?

R167:
> In typecheck and codegen the unhandled cases already default to `unhandledTag` which throws an exception. Should we be doing something more there?

2¢: I would advocate an explicit `case "bracket-lookup": throw new Error("Bracket lookup not implemented")` so that it's clear where/that this will be added in the future.

The reasoning behind the default `unhandledTag` is to make sure things can't fail silently on accident. Since `bracket-lookup` is intentional and being generated now by the parser, it should probably be explicitly handled (as opposed to before when there was no way to parse/generate it, so it would be an unexpected case).

lite-david:
@R167 I think I see your point, and probably that's what @jpolitz was going for as well. Let me make those changes. Thanks!

R167:
Looks like this is superseded by #61? Close this PR then?

lite-david:
Closing this since we've a new PR #61 with parsing, codegen, typechecking and tests

sudharshan-chakra:
Closing this PR as agreed in https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/pull/36#issuecomment-791633702

PR #37, opened 2021-02-28
Unify assignment and initial destructuring implementation

R167:
For destructuring, proceed on unifying assignment operations.
- Implement `assign` and `field-assign` in `assignment`
- Remove old-style assign code
- Add documentation on how `tcDestructure` and `codeGenDestructure` work.

This work is in further preparation for supporting full destructured assignment.

CC @ChadHayen @Kyle-Stadelmann


R167:
Note: this removed a decent amount of code duplication between tcStmt and tcExpr versions of variable/field assign/lookup, which was why I wanted to have the `Assignable<A>` expressions be a subset of `Expr<A>` union (rather than the current implementation of duplicating and needing to check for valid conversion)

jpolitz (approved):
LGTM

PR #40, opened 2021-03-01
Parsing for string literal, indexing & slicing

suhas-goutham:
Finished parsing for string literal, indexing & slicing.


sudharshan-chakra:
@suhas-goutham Hey, I'm from the dict team. Looks like you folks are using MemberExpression as well. There might be a conflict. Can we work on this to accommodate both cases?
@lite-david @hemathota @dheeraj7596

lite-david:
The following expr: ` s[10]` is going to be a `bracket-lookup` for us as well as the list folks. We decided on it [here](https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/pull/9#discussion_r583839792). Correct me if I'm wrong, but your changes in this PR would parse it as a your first case of `string_slicing`?  

CC: list folks as well @erauyeung @ricky8554

suhas-goutham:
> @suhas-goutham Hey, I'm from the dict team. Looks like you folks are using the case `"MemberExpression"` in `parser.ts` as well. There might be a conflict. Can we work on this to accommodate both of our cases?
> @lite-david @hemathota @dheeraj7596
@sudharshan-chakra 
Hey,
Yes for sure. We can work on it to accommodate both our cases.

suhas-goutham:
> @suhas-goutham The following expr: ` s[10]` is going to be a `bracket-lookup` for us as well as the list folks. We decided on it [here](https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/pull/9#discussion_r583839792). Correct me if I'm wrong, but your changes in this PR would parse it as a your first case of `string_slicing`?
> 
> Also we have a PR for `bracket-lookup` parsing [here](https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/pull/36)
> 
> CC: list folks as well @erauyeung @ricky8554

@lite-david 
Yes, you're right. It would pass it as the first case of string slicing.
Also, would it be okay with you guys if our group would make changes in our parser to return "bracket-lookup" in our first case instead of us returning "string-slicing" in our first case. I believe this would accommodate  both our cases.

The code would look like this:-

`if(sliced_list[0]!=""){
          start_index = traverseExpr(c, s);
          console.log("First case "+s.substring(c.from, c.to));
          //objExpr would be "s"  and start_index would be 10 in the example "s[10]"
          if(sliced_list.length==1){
            return {  tag: "bracket-lookup", obj: objExpr, key: start_index }
          }
          c.nextSibling();
        }`

Please let me know what you think about this.

CC: Lists group @ricky8554 I believe this parsing would work for you guys as well. We can rename the AST to just "slicing" instead of "string_slicing" if you guys are also willing to use the parsing we have done here. Can you check out this PR and let us know what you think about it ?

jpolitz (changes requested):
I would rather not merge untested implementation code. Can we get some tests associated with this, and some kind of “not yet implemented” exception in the places where the new syntax would need to be handled in other parts of the compiler?

jpolitz:
**Thanks** for submitting an incremental pull request!

suhas-goutham:
> I would rather not merge untested implementation code. Can we get some tests associated with this, and some kind of “not yet implemented” exception in the places where the new syntax would need to be handled in other parts of the compiler?

@jpolitz Sure, we will do that. I just wanted to push the parsing idea first since many groups would have conflicting cases here and we could discuss the issue. So, we had just opened a Pull Request for parsing.
We will implement the feature with tests and submit the PR.

compiler.ts
```diff
@@ -305,15 +322,106 @@ function codeGenExpr(expr: Expr<Type>, env: GlobalEnv): Array<string> {
...
+        var brKeyStmts = codeGenExpr(expr.key, env);
+        var brStmts = [];
+        //First check whether index is out of bounds, if so, throw error by going to print_str. Then index the string.
+        brStmts.push(
+          ...[
+            `${brObjStmts.join("\n")}`, //Load the string object to be indexed
+            `(local.set $$string_index)`,
+            `${brKeyStmts.join("\n")}`, //Add the index * 4 value to the address
+            `(local.get $$string_index)(i32.load)(i32.gt_s)`,
+            `(if (then (i32.const -1)(call $print_str)(drop)))`, //Check if string index is out of bounds
```
jpolitz, line 337: @h9jiang error reporting case – what should they be doing here?

compiler.ts
```diff
@@ -305,15 +322,106 @@ function codeGenExpr(expr: Expr<Type>, env: GlobalEnv): Array<string> {
...
+  var i = 1;
+  //Storing the length of the string at the beginning
+  stmts.push(
+    ...[
+      `(i32.load (i32.const 0))`, // Load the dynamic heap head offset
+      `(i32.const ${string_val.length - 1})`, // Store ASCII value for 0 (end of string)
+      "(i32.store)", // Store the ASCII value 0 in the new address
+    ]
+  );
+  while (i != string_val.length + 1) {
```
jpolitz, line 389: This makes sense, but I wonder if there is some better way to set an array of memory in WASM rather than 1 byte at a time.

compiler.ts
```diff
@@ -305,15 +322,106 @@ function codeGenExpr(expr: Expr<Type>, env: GlobalEnv): Array<string> {
...
+      ...[
+        `(i32.load (i32.const 0))`, // Load the dynamic heap head offset
+        `(i32.add (i32.const ${i * 4}))`, // Calc string index offset from heap offset
+        `(i32.const ${char_ascii})`, // Store the ASCII value of the string index
+        "(i32.store)", // Store the ASCII value in the new address
+      ]
+    );
+    i += 1;
+  }
+  //At end of string, we store ASCII value 0 which represents null
```
jpolitz, line 401: Do we need this if we store the length?

parser.ts
```diff
@@ -188,15 +196,70 @@ export function traverseExpr(c: TreeCursor, s: string): Expr<null> {
...
-      c.nextSibling(); // Focus on .
-      c.nextSibling(); // Focus on property
-      var propName = s.substring(c.from, c.to);
-      c.parent();
-      return {
-        tag: "lookup",
-        obj: objExpr,
-        field: propName,
-      };
+      c.nextSibling(); // Focus on . or [
```
jpolitz, line 199: @ricky8554 @erauyeung can you compare this to your bracket lookup solution and see what a good harmonized solution is?

tests/import-object.test.ts
```diff
@@ -1,10 +1,24 @@
 import { Type } from "../ast";
-import { NUM, BOOL, NONE, unhandledTag } from "../utils";
+import { NUM, STRING, BOOL, NONE, unhandledTag } from "../utils";
 
 function stringify(typ: Type, arg: any): string {
   switch (typ.tag) {
     case "number":
       return (arg as number).toString();
+    case "string":
```
jpolitz, line 8: Nice!

tests/import-object.test.ts
```diff
@@ -1,10 +1,24 @@
...
+      if (arg == -1) throw new Error("String index out of bounds");
```
jpolitz, line 9: Ah I see. @h9jiang FYI. I would make this a separate function rather than re-using the same one to report the error and print.

tests/import-object.test.ts
```diff
@@ -22,6 +36,11 @@ function print(typ: Type, arg: any): any {
   return arg;
 }
 
+const memory = new WebAssembly.Memory({ initial: 2000, maximum: 2000 });
```
jpolitz, line 39: @student-of @SasyaReddi note the need to create the memory in tests and in webstart so the printing functions can close over it. May need to think about the right place to do this. Also @lsteffkojr talked about aggregating some of the shared import stuff in a library, relevant here (and something similar needed for allocated bignums).

jpolitz (commented):
Leaving as a comment because I want to give feedback, but wait on the parsing of bracket-lookup for now.

Since slicing isn't used yet (just parsed) maybe we put that off and work with an existing bracket-lookup impl

compiler.ts
suhas-goutham: Yes, you're right. We don't need it now that we put string length at the beginning. Thanks!

suhas-goutham:
> Leaving as a comment because I want to give feedback, but wait on the parsing of bracket-lookup for now.
> 
> Since slicing isn't used yet (just parsed) maybe we put that off and work with an existing bracket-lookup impl

Yes sure. It looks like there is no bracket lookup merged in the main branch. We will use the bracket-lookup parsing similar to the one done by the lists group for this milestone. @ricky8554 @erauyeung But we have checked the parsing which is done for slicing by us and it works correctly. We also feel that the lists group can benefit from using the same "slicing" parsing which we have done since strings and lists have same slice operations. Our parsing for slicing also supports the simple bracket-lookup case. If you check out our first case in slicing, we return the {tag:"bracket-lookup"}

tests/import-object.test.ts
suhas-goutham: @jpolitz We are unsure about this comment. Can you please clarify what the new function must be doing? For the case, arg == -1, should we call a new function where we just throw an error or should we wait for another group to implement this error handling?

type-check.ts
```diff
@@ -413,6 +413,19 @@ export function tcExpr(env: GlobalTypeEnv, locals: LocalTypeEnv, expr: Expr<null
...
         throw new TypeCheckError("method calls require an object");
       }
+    case "bracket-lookup":
+      var obj_t = tcExpr(env, locals, expr.obj);
+      var key_t = tcExpr(env, locals, expr.key);
+      var tBracketExpr = { ...expr, obj: obj_t, key: key_t, a: obj_t.a };
+      if (obj_t.a != STRING) {
+        throw new TypeCheckError("Bracket lookup on " + obj_t.a.tag + " type not possible");
+      }
+      if (key_t.a != NUM) {
```
R167, line 423: You might want to use a type comparison helper function here.

jpolitz (approved):
Good enough! Lots of useful stuff that we need for making progress, but we'll come back to these comments :-)

tests/import-object.test.ts
lsteffkojr: I just created a new pull request for this idea of aggregating the import stuff.

compiler.ts
```diff
@@ -261,26 +269,35 @@ function codeGenExpr(expr: Expr<Type>, env: GlobalEnv): Array<string> {
       return valStmts;
     case "construct":
       var stmts: Array<string> = [];
+      stmts.push(
+        ...[
+          "(i32.const 0)", // Address for our upcoming store instruction
+          "(i32.load (i32.const 0))", // Load the dynamic heap head offset
+          "(local.set $$string_class)",
+          "(i32.load (i32.const 0))",
+          `(i32.add (i32.const ${env.classes.get(expr.name).size * 4}))`, // Move heap head beyond the two words we just created for fields
```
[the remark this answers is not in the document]
suhas-goutham: Thanks! Pushed that change in the new PR.

type-check.ts
suhas-goutham: Thanks for the suggestion!

PR #41, opened 2021-03-02
Big nums literal tags

actomlinson:
Tagging is required to implement big numbers efficiently, we can tag integers < 31 bits as literals and promote integers > 31 bits to objects. It is also required to implement garbage collection. This is the big num proposal to implement tagging.

~~Since literals have a 1 in the LSB, the typescript min/max/abs/pow functions no longer return the correct result. This can be fixed by writing new functions that decode the result. In this version though that is not implemented so those tests are broken.~~

These tests are working again in this latest commit, which decodes literals before applying the built in functions. This approach will need to be changed some time in the future once we want built ins to work on big nums.

@Link-lin @lsteffkojr @student-of @SasyaReddi


jpolitz:
Just a note that I merged over some autoformatting stuff that's causing the conflicts you see.

jpolitz:
Nice! This will usefully break a bunch of folks' code (like bracket lookup on lists). I think I will merge it _last_, so we can see the working state and then see what this breaks.

jpolitz:
How does this relate to the milestone for bignums? Is it a prereq, overridden by, etc?

lsteffkojr:
This is a subset of the bignums milestone that addresses the LSB tagging scheme for literals. We prepared this earlier in the week in case any discussion was needed on how this might affect other projects. The milestone PR now includes these same code changes that are in this PR.

jpolitz:
Closing as included in the other milestone PR.

PR #44, opened 2021-03-02
Comprehensions

thecooltechguy:
Our initial 2 example programs are working, and we've finished our proposal.


ast.ts
```diff
@@ -100,7 +100,14 @@ export type Expr<A> =
...
   | { a?: A; tag: "lambda"; args: Array<string>; ret: Expr<A> }
-  | { a?: A; tag: "comprehension"; expr: Expr<A>; field: string; iter: Expr<A>; cond?: Expr<A> }
+  | {
+      a?: A;
+      tag: "comprehension";
+      expr: Expr<A>;
+      field: Assignable<A>;
+      iter: Expr<A>;
+      cond?: Expr<A>;
+    } // Need to change field to Assignable since lst = [t.n for t.n in range(10)] runs correctly, if the type of t has n.
```
R167, line 110: It might be worth using `Destructure<A>` (or `tag: "assignment"`) over `Assignable<A>` since python allows destructuring in this position (e.g. `[x for x, _ in [(1, 2)]]`), in particular because the `(tc|traverse|codeGen)Destructure` are the endpoints we plan to support for general purpose assignment. See #37 for proposed [API](https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/pull/37/files#diff-0a03798f0e47b1080516ea97c55cfc6c2e025276ef50bf1e1ede9b6e4570a33cR225-R238).

runner.ts
```diff
@@ -49,6 +49,39 @@ export async function run(
   source: string,
   config: Config
 ): Promise<[Value, compiler.GlobalEnv, GlobalTypeEnv, string]> {
+  // One approach to create built-in Range class and range function, by Comprehension team
+  if (!config.typeEnv.classes.has("Range")) {
```
jpolitz, line 53: Is there a way to do this without concatenating code?

tests/comprehensions.test.ts
```diff
@@ -0,0 +1,25 @@
+import { assertPrint } from "./utils.test";
+
+describe("comprehensions test", () => {
```
jpolitz, line 3: Thanks for adding these! It's great that the tests from the proposal work. There are some more tests I think are important to write – the ones from your proposal are representative, but not thorough, tests.

Right now nothing uses the expression position – what about `[i * 2 for i in range(0, 5)]`?

What about creating multiple comprehensions in the same program? That would make sure nothing is global.

What about using the result of a comprehension in another comprehension? For example

```
a = [i * 2 for i in range(0, 5)]
b = [i + 1 for i in a]
```

These seem like they would be good tests of the _existing_ infrastructure that work in Python and should work with what's implemented so far.

Or what about `[print(i) for i in range(0, 5)]`, which would demonstrate that the expression is called?

transform.ts
```diff
@@ -0,0 +1,45 @@
+import { Expr, Type } from "./ast";
+
+export function transformComprehension(comprehension: Expr<Type>): Expr<Type> {
+  if (comprehension.tag !== "comprehension") {
+    throw new Error("Expected a comprehension expression");
+  }
+
+  // TODO: Fully implement this function once we have some of the other parts of the compiler working
+  // Ensure iter: is a function call to range
+  if (comprehension.iter.tag !== "call" || comprehension.iter.name !== "range") {
```
jpolitz, line 10: Good hack to get started. I wonder if you could give a type to comprehensions and let this be more general, and support other values next.

jpolitz (changes requested):
It's not clear to me if the _expression_ position of comprehensions runs. I think that's an important thing to have tests for and should work with the constraints you have (e.g. no lists yet, etc).

ast.ts
thecooltechguy: This is a great suggestion @R167 . 

We were thinking of putting in `Destructure<A>` once the proposed API is merged, which doesn't seem to be the case right now. But assuming this gets merged before the next milestone, we are planning to make the change that you suggest.

thecooltechguy:
> It's not clear to me if the _expression_ position of comprehensions runs. I think that's an important thing to have tests for and should work with the constraints you have (e.g. no lists yet, etc).

Yes, in our current milestone version, the _expression_ position of comprehensions doesn't run. 

The reason for this is because,  even if we run the _expression_ position for all of the elements in the comprehension, we won't be able to collect the output results back into a `list` container (and we can't really use our mock `range` class to collect the output results either).

So, our idea was to first get this milestone version working for only supporting use-cases where the expression is just the element in our range object, and once list support is added, we were planning to change this so that the expression component is actually evaluated and all the results are collected into a list output.

tests/comprehensions.test.ts
thecooltechguy: Oh ok, that makes sense @jpolitz . 

We've currently added some tests to also account for this expression part of comprehensions, but in our current implementation version, this fails (since comprehension expressions aren't currently being evaluated, as explained in my other comment).

So, once we integrate list support into comprehensions, these tests related to the evaluation of comprehension expressions should begin to pass at that stage.

runner.ts
thecooltechguy: @jpolitz Currently, we followed this (rather unclean) approach, since our current implementation is kind of a temporary "hack" and only supports `range(...)` calls as iterators. 

But our ideal end goal is to replace all this by supporting lists natively, without the need for any sort of built-in functions/classes, like `range(...)` and `Range`.

So we were thinking that, at that point, we would just remove this part here.

transform.ts
thecooltechguy: Oh ok, so maybe instead of `comprehension.iter` being an `Expr<A>`, we can instead have it be something like `ComprehensionIterable<A>`, where this `ComprehensionIterable` type can be a list, etc.

thecooltechguy:
Hi @jpolitz , we've updated this PR with our final project's most up to date progress (instead of creating a new separate PR).

Currently, we manually checked our features to ensure that they work, but the tests are failing due to some bugs in the for loop/lists implementations. We expect these to pass once the latest PRs from the for loop and list teams are merged.

Issue #43, opened 2021-03-02
"while" loop is compiled as a "do while" loop


ALMSIVI:
For the following program:
```python
while False:
  print(1)
```

the compiler generates the following wasm code:
```wat
(block (loop (i32.const 1)
  (call $print_num)
  (local.set $$last)
  (br_if 0 (i32.const 0))
  (br 1)
))
```

The wasm code doesn't check the condition at the beginning of the loop; rather, it will execute the body first before checking the condition; this essentially becomes a `do...while` loop, and for the above program, `1` is incorrectly printed to the console.

PR #45, opened 2021-03-02
Fixed `while` issue

ALMSIVI:
Fixed the while loop compiler issue mentioned in #43


R167:
Yay! Maybe also adding a regression test for the case you mentioned in #43, just as a precaution for the future and further proof this works(?) (e.g. `assertPrint("while-eval-beginning", "while False: print(0)", "")`)

ALMSIVI:
Added a test in `runner.test.ts`.

jpolitz (approved):
LGTM

PR #47, opened 2021-03-03
Import statement parser code and built-in libraries proposal

TypeMonkey:
This pull request adds parser code for import statements of the form:

`import <moduleName>`  and
`from <moduleName> import <moduleComponent>` where moduleComponent is a class, function or file variable

It also includes a proposal for general imports for libraries.


TypeMonkey:
I've submitted a new pull request with an updated proposal.

PR #49, opened 2021-03-04
Milestone error-reporting PR

h9jiang:
**Changes in this PR:**

1. We added more details and changed constructors of the error classes in error.ts. The reference of our error classes: https://docs.python.org/3/library/exceptions.html.

2. One thing to note: These error classes contain dynamic runtime error classes that will be implemented by March 11. 

3. TypeCheckError and TypeError in type-check.ts are removed. Please throw corresponding errors using BaseException from error.ts.

4. The attribute “loc” is of type Location, which is handled in `parser.ts`, and is used to track the position of each fragment of the code. 

    We added three new customized error classes, given that some errors will not be thrown in real python: 
    4.a UnsupportedOprandTypeError -> This error class is for TypeError related to type mismatch around operators like + - // * ....
    4.b TypeMismatchError -> This error class is for TypeError that is allowed in Python. For example, `a:int = False` is allowed in python.
    4.c ConditionTypeError -> This error class is for condition type check in while and if. For example, “if 1” is allowed in python.

**Next steps:**
Dynamic error throwings, stacktrace. Details are included in `milestone-error-reporting.md`


error.ts
```diff
@@ -44,51 +50,57 @@ export class KeyboardInterrupt extends Error {
...
-
     // Alternatively use Object.setPrototypeOf if you have an ES6 environment.
     this.__proto__ = trueProto;
   }
 }
 
 export class StopIteration extends Exception {
-  constructor(message?: string, name = "StopIteration") {
-    super(message, name);
+  constructor(message?: string, loc?: Location) {
```
jpolitz, line 71: Do you anticipate having lots of cases where `loc` is in fact not provided to the constructors?

ast.ts
```diff
@@ -83,35 +94,58 @@ export const ASSIGNABLE_TAGS = ["id", "lookup"] as const;
  * Subset of Expr types which are valid as assign targets
  */
 export type Assignable<A> =
```
jpolitz, line 96: Is there any way we could use the `a` field for the `Locations`? For example:

```
parse : string -> Expr<Location>
type-check: Expr<Location> -> Expr<[Type, Location]>
```

We could add new elements to the tuple each time.

Then we could do e.g. `{ a: [typ, ...expr.a], [ more fields here ]}` in the type checker or other similar contexts that just add a single element to the front. We could also do it with an object type in that position. But it seems like a lot to always carry around the location, and to have _another_ optional field that we could forget in many positions.

error.ts
```diff
@@ -97,135 +109,169 @@ export class ZeroDivisionError extends ArithmeticError {
...
-      Error.captureStackTrace(this, ZeroDivisionError);
+      Error.captureStackTrace(this, TypeError);
+    }
+  }
+}
+
+export class TypeMismatchError extends Exception {
+  constructor(expect: string, got: string, loc?: Location) {
+    super(`Expected type '${expect}'; got type '${got}'`, "TypeMismatchError", loc);
+    if (Error.captureStackTrace) {
```
jpolitz, line 229: I think we should distinguish _static errors_ and _dynamic errors_, because a stack trace on a static error doesn't make much sense. Also, shouldn't these relate to BaseException somehow?

There's a lot of duplicated code here. What does `if (Error.captureStackTrace)` mean? I'm generally a little confused about these patterns. Maybe I missed a link to some discussion about why things are don this way?

jpolitz: I also think we might want to save the values of _all_ the fields in the error. That way a fancier front-end rendering could have access to those values. and in a type mismatch, it's probably useful to keep around the whole Type rather than pre-converting it to a string before calling the constructor (since the front-end might want to distinguish that as well, especially with e.g. list types that are compound structures).

jpolitz (changes requested):
I'd suggest more clearly differentiating static and dynamic errors.

I'd like to see fields stored on the errors with richer information than pre-stringifying, because the front-end can use more information to print things well.

I'm curious for your thoughts on using the `a` position in the AST rather than adding a new optional `loc` field everywhere.

Also curious for your thoughts on the code structure questions I have about stack traces and duplicated code in the error classes.

error.ts
FrancisBian: Because at the beginning we assumed every group would merge their implementation during this milestone. So we planned to make it optional and then help other groups add location in parser.ts after complete merge.
We are making it required and also move it into the "a" tag.

ast.ts
FrancisBian: Will do. 

Our initial though on adding location as another field was because we thought the location does not need to go through the decorating process. But you are right, it is more efficient and extendable to use the "a" tag. (We forgot to talk about this during your OH yesterday, since you mentioned this in lecture before).

error.ts
FrancisBian: Editing...

FrancisBian:
Do you have any suggestions on changing the structure to differentiae static and dynamic errors? 
Our thoughts right now based on your comment is that:

We think NameError,  AttributeError, SyntaxError, TypeError, and the four new Errors (TypeMismatch, UnsupportedOpperand, ConditionTypeError) are static. So they can extends a super class called StaticError. We don't need to have stacktrace in those classes (Error class in TypeScript has an optional "stack?").

All other errors are dynamic errors. They can extend a super class called DynamicError. And we need to figure out how to get reasonable stacktrace from them.

Am I on the right track?

error.ts
jpolitz: Got it, good idea.

jpolitz:
Yeah I think a simple hierarchy works here. Since TS supports unions, we could even have them disjoint.

`runner.run` could throw either since it compiles _and_ runs. But it's fine for it to throw `CompileError | RuntimeError`

FrancisBian:
We modified the code structure and added richer fields. Next steps would be dynamic errors and stack trace, after merged.

FrancisBian:
We are updating tests for last commit. Will pass all checks soon.

FrancisBian:
Done solving existed merge conflicts

jpolitz (approved):
Wow, I'm impressed you kept up with all the changes. I'm surprised at some of the reformatting in ast.ts, but other than that LGTM.

PR #50, opened 2021-03-04
Milestone-Lists PR

ricky8554:
@erauyeung 

2 of our programs works in this milestone. New proposal was added to the proposal directory.


parser.ts
```diff
@@ -398,9 +456,32 @@ export function traverseStmt(c: TreeCursor, s: string): Stmt<null> {
...
+      "Can Not Parse Type " + s.substring(c.from, c.to) + " " + c.node.from + " " + c.node.to
+    );
+  }
+}
+
 export function traverseType(c: TreeCursor, s: string): Type {
   // For now, always a VariableName
+
   let name = s.substring(c.from, c.to);
+  if (name.includes("[")) return traverseBracketType(c, s);
```
jpolitz, line 484: This seems brittle. Is there a specific tag name or test we can do in order to detect this? I see that `traverseBracketType` goes into children, so presumably it has a recognizable name other than substring-ing?

proposals/milestone-lists.md
```diff
@@ -0,0 +1,111 @@
...
+n = items[0] + items[1]
+```
+
+![](https://i.imgur.com/9UPlFXJ.png)
+
+
+### 2. A description of which examples will work by March 11, including any updates you want to make to the examples you plan for March 11
+
+#### Program 1
+We are scaling back our Program 1 because inheritance may be too much to handle in the time we have left.
```
jpolitz, line 29: Yeah this isn't really your job anyway.

proposals/milestone-lists.md
```diff
@@ -0,0 +1,111 @@
...
+items = [1, 2, 3] + [4, 5, 6]
+```
+
+#### Program 5
+```python
+items : [int] = None
+items = [1, 2, 3] 
+items[1] = 90
+```
+#### Program 6 &#8594; Runtime error (Index out of bounds, error code 3)
```
jpolitz, line 64: Is this error coordinated at all with the error-reporting group? @h9jiang

type-check.ts
```diff
@@ -59,16 +59,20 @@ export type TypeError = {
   message: string;
 };
 
-export function equalType(t1: Type, t2: Type) {
-  return t1 === t2 || (t1.tag === "class" && t2.tag === "class" && t1.name === t2.name);
+export function equalType(t1: Type, t2: Type): boolean {
+  return (
+    t1 === t2 ||
+    (t1.tag === "class" && t2.tag === "class" && t1.name === t2.name) ||
+    (t1.tag === "list" && t2.tag === "list" && equalType(t1.content_type, t2.content_type))
```
jpolitz, line 66: 👍

type-check.ts
```diff
@@ -59,16 +59,20 @@ export type TypeError = {
...
+  );
 }
 
 export function isNoneOrClass(t: Type) {
   return t.tag === "none" || t.tag === "class";
 }
 
 export function isSubtype(env: GlobalTypeEnv, t1: Type, t2: Type): boolean {
-  return equalType(t1, t2) || (t1.tag === "none" && t2.tag === "class");
+  return equalType(t1, t2) || (t1.tag === "none" && (t2.tag === "class" || t2.tag === "list"));
```
jpolitz, line 75: 👍

type-check.ts
```diff
@@ -381,6 +388,18 @@ export function tcExpr(env: GlobalTypeEnv, locals: LocalTypeEnv, expr: Expr<null
       } else {
         throw new TypeCheckError("field lookups require an object");
       }
+    case "bracket-lookup":
+      var tObj = tcExpr(env, locals, expr.obj);
+      var tKey = tcExpr(env, locals, expr.key);
+      if (tObj.a.tag === "list") {
+        if (tKey.a.tag === "number") {
+          return { ...expr, a: tObj.a.content_type, obj: tObj, key: tKey };
```
jpolitz, line 396: 👍

type-check.ts
```diff
@@ -413,6 +432,29 @@ export function tcExpr(env: GlobalTypeEnv, locals: LocalTypeEnv, expr: Expr<null
...
       }
+    case "list-expr":
+      var commonType = null;
+      const listExpr = expr.contents.map((content) => tcExpr(env, locals, content));
+      if (listExpr.length == 0) {
+        commonType = NONE;
+      } else {
+        commonType = listExpr[0].a;
+        for (var i = 1; i < listExpr.length; ++i) {
+          //takecare of nested list
```
jpolitz, line 443: I'm confused by this comment.

utils.ts
```diff
@@ -10,6 +10,8 @@ export function PyValue(typ: Type, result: number): Value {
       return PyObj(typ.name, result);
     case "none":
       return PyNone();
+    case "list":
+      return PyObj(typ.tag + `<${typ.content_type.tag}>`, result);
```
jpolitz, line 14: 👍

tests/lists.test.ts
```diff
@@ -1,15 +1,47 @@
...
+        class A(object):
+            n:int = 100
+        x : [A] = None
+        x1 : A = None
+        x2 : A = None
+        x3 : A = None  
+        x = [A(),A(),A()]
+        `;
+
+  assert("List With Class", source, PyNone());
```
jpolitz, line 46: This seems like a weak test, in that it doesn't say anything about the _contents_ of the list. Maybe try some list access along with this list creation to make sure the objects are accessible?

tests/lists.test.ts
```diff
@@ -1,15 +1,47 @@
...
+        items = [1, 2]
+        n = items[0] + items[1]
+        n
+    `;
+
+  assert("Lists Access (Program 4)", source, PyInt(1 + 2));
+
+  //Other Test
+  assert("Empty List", "[]", PyObj(`list<none>`, 540));
+  assert("List With Number", "[1,2,3]", PyObj(`list<number>`, 632));
```
jpolitz, line 32: I think these tests are brittle in the presence of any memory management/changing allocation strategies. Maybe it would be best to find a way to check just the list type part, and let the address be anything?

tests/lists.test.ts
```diff
@@ -1,15 +1,47 @@
...
+  var source = `
+        items : [int] = None
+        items = [1, 2, 3]
+    `;
+
+  assert("Assign List To Variable (Program 2)", source, PyNone());
+
+  var source = `
+    items : [int] = None
+    items = [1, 2, 3] + [4, 5, 6]
```
jpolitz, line 14: Would like to see a version of this test where the two lists are stored in variables, then concatenated.

tests/lists.test.ts
```diff
@@ -1,15 +1,47 @@
...
-  //     n
-  // `
-  // assert('Lists Access', source,  PyInt(1 + 2))
+describe("LIST TEST", () => {
+  var source = `
+        items : [int] = None
+        items = [1, 2, 3]
+    `;
+
+  assert("Assign List To Variable (Program 2)", source, PyNone());
```
jpolitz, line 10: Weak test – it just says there was no type error. What do we know about `items`?

jpolitz (changes requested):
Nice work.

Some comments about the code in the compiler and parser that are worth addressing.

The main thing to add is that there need to be more, and better, tests of these same features. It's great that these work, but let's explore what we can do with list literals, bracket lookup, and concatenation to make sure they are solid.

parser.ts
erauyeung: We're changing this now, we can use `ArrayExpression` to recognize things like `[int]` as the type.

compiler.ts
```diff
@@ -305,6 +496,66 @@ function codeGenExpr(expr: Expr<Type>, env: GlobalEnv): Array<string> {
       var className = objTyp.name;
       var [offset, _] = env.classes.get(className).get(expr.field);
       return [...objStmts, `(i32.add (i32.const ${offset * 4}))`, `(i32.load)`];
+    case "bracket-lookup":
+      var objStmts = codeGenExpr(expr.obj, env);
+      //This should eval to a number
+      //Multiply it by 4 to use as offset in memory
+      var keyStmts = codeGenExpr(expr.key, env);
```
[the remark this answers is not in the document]
lsteffkojr: Yes, this would be problematic because the result of evaluating the keyStmts will have a tag bit in the LSB. Our pull request will provide a helper that removes the tag bit

compiler.ts
```diff
@@ -244,6 +433,8 @@ function codeGenExpr(expr: Expr<Type>, env: GlobalEnv): Array<string> {
     case "binop":
       const lhsStmts = codeGenExpr(expr.left, env);
       const rhsStmts = codeGenExpr(expr.right, env);
+      if (typeof expr.left.a !== "undefined" && expr.left.a.tag === "list")
+        return [...rhsStmts, ...lhsStmts, ...codeGenListCopy(true), ...codeGenListConcat()];
```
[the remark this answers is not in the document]
ricky8554: We are planning to add copy method for lists which will be implemented later this week. The codeGenListCopy function will return code for list copy if the concat parameter is set to false.

compiler.ts
```diff
@@ -239,7 +243,7 @@ function codeGenAssignable(target: Assignable<Type>, value: string[], env: Globa
     default:
       // Force type error if assignable is added without implementation
       // At the very least, there should be a stub
-      const err: never = target;
+      const err: never = <never>target;
```
R167, line 246: Note for future:
This change invalidated the entire check that was being done here. doing this forcibly typecast the target to `never`, avoiding the entire point of the check requiring a stub for `bracket-lookup` not being implemented. It should have instead been resolved with a `case "bracket-lookup": throw new Error("Assignment to bracket lookup not yet supported")`

.gitignore
```diff
@@ -3,3 +3,4 @@ build
 cli
 *.sw*
 mochawesome-report
+*.js
```
R167, line 6: Note: This likely should have been more targeted since we already have several JS files (correctly) checked in and could potentially cause headaches if trying to include a file like `.eslintrc.js` in the future:
```
cypress/integration/frontend.spec.js
cypress/plugins/index.js
cypress/support/commands.js
cypress/support/index.js
parse-python.js
webpack.config.js
```
fortunately git will continue tracking changes to these existing files.

PR #51, opened 2021-03-04
Milestone for fancy calling conventions

danielytan:
For the milestone, we implemented support for basic default parameters, which are implemented primarily in the typechecker. In the parser, we enforce that default parameters must be the rightmost parameters. In the typechecker, we append default values for missing arguments that have corresponding default parameters.


parser.ts
```diff
@@ -427,12 +428,17 @@ export function traverseParameters(c: TreeCursor, s: string): Array<Parameter<nu
     let typ = traverseType(c, s);
     c.parent();
     c.nextSibling(); // Move on to comma or ")" or "="
-    nextTagName = c.type.name; // NOTE(daniel): copying joe's hack for now
+    nextTagName = c.type.name; // NOTE(daniel): copying joe's hack for now (what would be the proper way to avoid this?)
```
jpolitz, line 431: I would lift this code out into a helper function. This is getting long.

proposals/milestone-fancy-calling-conventions.md
```diff
@@ -0,0 +1,95 @@
...
+  print(add_default_10(20, 5))`, ['30', '25']);
+  
+  assertPrint("project-proposal program 2", `
+  def add_defaults(x : int = 10, y : int = 20, z : int = 30) -> int:
+	  return x + y + z
+
+  print(add_defaults())
+  print(add_defaults(40))`, ['60', '90']);
+```
+These programs differ slightly from the ones in our original proposal because verify correctness using print statements here. Again, program 1 tests for a single default parameter, and program 2 tests for multiple (i.e. all) default parameters.
```
jpolitz, line 19: Good idea to use print.

type-check.ts
```diff
@@ -351,14 +380,36 @@ export function tcExpr(env: GlobalTypeEnv, locals: LocalTypeEnv, expr: Expr<null
...
+          // first populate all the arguments first.
+          // Then, populate the rest of the values with defaults from params
+          var augArgs = expr.arguments;
+          var argNums = expr.arguments.length;
+          while (argNums < argTypes.length) {
+            if (params[argNums].value === undefined) {
+              throw new Error("Missing argument from call");
+            } else {
+              // add default values into arguments as an Expr
+              augArgs = augArgs.concat({ tag: "literal", value: params[argNums].value });
```
jpolitz, line 408: Interesting! So this is really restricted to literals. I think a totally different strategy will be needed for non-literals, because we can't copy those expressions around.

jpolitz (approved):
Overall approve, though there are a few recommendations I have here.

# Types

@DDEle and the closures group are going to be interested in how type-checking functions works if they have parameters. For example, if an argument is of type `Callable[int, int]`, can we subtype a function of type `Callable[int, int=5, int]` to that type? (Try to turn that into an example if you're not sure what I'm talking about).

# Restriction to Literals

The restriction to literals is a bit unfortunate. I'd like to push you a little bit to think about how you might handle this for other kinds of expressions. Let me make a suggestion.

You already have the infrastructure in the type checker to add the extra arguments for unfilled defaults (though it might be a little tricky to do in the first-class functions case I suggested above, let's set that aside for now).

We could make up a new tag for a value that means “uninitialized parameter.” Then, in the _body_ of the function, we could add statements at the beginning that check if a parameter is uninitialized, and if it is, assign the parameter to the correct initial value.

This would make the type-checker no longer add extra arguments drawn from the defaults, but instead always add an appropriate number of extra arguments with a default “uninitialized” value. Then the function body could detect this difference with a runtime tag check.

type-check.ts
xiexr151e: @DavidLiu0304 @danielytan 
CC @DDEle because Joe's suggestion completely reverses the restriction what we decided earlier.

I would like some clarification on your suggestion above.
So, our goal is to enable not just literals, but expressions for default arguments. Let's use an example for this:
```
class Bar(object):
    baz : int = 0

def f(foo : int, bar : Bar()) -> int:
    return foo + bar.baz

print(f(7))
```

If I'm understanding this correctly, in the above example, `foo` would be the aforementioned uninitialized parameter. `foo` will be tagged with the hypothetical "uninitialized parameter" tag. This would just be adding a new tag to the AST, right?
```
We could make up a new tag for a value that means “uninitialized parameter.” Then, in the body of the function, we could add statements at the beginning that check if a parameter is uninitialized, and if it is, assign the parameter to the correct initial value.
```
I'm not sure what this means. Above, `foo` is uninitialized, but how would we assign the parameter to the correct initial value? In the compiler, don't we already do that just by putting arguments on the stack and make the call? What about overriding a default value by providing an argument to a parameter?

DavidLiu0304: From what I envision, in compiler.ts, when we encounter an "uninitialized parameter" in a function call, we would generate the code for that parameter much like the code generation for "construct". Then, we could use that value in the argument for the function call.

xiexr151e: @jpolitz 

I had some discussion with @DavidLiu0304 and @danielytan again, and we just realized that we don't have to be just restricted to literals for default arguments. 
One point I brought up during our meeting is that, in the parser, if we change traversing for literal to traversing for expression, then can't we just use the existing methodology to also do something like:
```
def func(a : foo = Foo()):
  return a.x
func()
```
In other words, in the existing type check, all we have to do is just insert a `construct` expression.
I believed that we previously were not able to do this because we had this restriction in the parser.

jpolitz: Expressions in general are interesting here.

If the constructor `Foo()` prints, how many times should we see a print happen if we call `func` 10 times?

Issue #18, opened 2021-02-25
If statement return type bug


R167:
Comparison of the return types of if statements currently has an issue that it uses exact object reference equality to compare return types. This is generally fine for simple types like bool, num, and none, since we use constants to define them. However, This becomes problematic for object style return types since the `CLASS(name: string)` helper always creates a new JS object for a type. The below line is the problem.

https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/blob/7fe2ed68619697dc7b8d757812bb45f894122adc/type-check.ts#L110-L111

```python
class blah():
  x: int = 0

def test() -> blah:
  if True:
    return blah()
  else:
    return blah()
test().x
```

Expected: `0`
Got: "Error: Types of then and else branches must match"

This issue may best be handled by the inheritance group (#11) or lists group (#9) since they'll be working on extending the core behaviour to check if two types are joinable.

Finally, there is a second, related issue here. When determining the return value of a block, we naively take the type of the last value, however this is not necessarily the actual return statement. For example, while somewhat odd, this code is perfectly legal, but meets a tc error (legal because we don't specify `return` must be at the end of a block).

```python
def test() -> bool:
  if True:
    return True
    8
  else:
    return False
test()
```
Expected: `True`
Got: "Error: Types of then and else branches must match" (this is because it compares type of `8` with `return False`).

PR #53, opened 2021-03-04
Milestone for Inheritance

Freyr-Wings:


Freyr-Wings:
In the milestone, we add `test/inheritance.test.ts`, which includes the tests mentioned in our proposal and some tests for issue #18.

jpolitz:
This is super cool! I've done a broad overview – are there specific areas you would like feedback or advice on?

PR #62, opened 2021-03-05
Milestone for Closures

DDEle:
Here's a brief summary of changes:

- Some changes of `Closure` structure in AST.
- Add a `call_expr` expression, whose callee is an expression. The vanilla call expression should not be used by the parser.
- Add an "escape analysis" phase between tc and codegen, which currently assumes that every function escapes.
- Support in `parser`, `type-check`, `compiler` file.


DDEle:
We fix the conflict (regarding destructure & string?) last night and passed all the test cases. However, we are not sure how to resolve conflicts now (after fancy-calling-conventions (FCC) is merged) as there is a nontrivial design decision about how type-checking functions work if they have default parameters and more problems derived from that in terms of `ea` and `codegen`, as mention by @jpolitz in https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/pull/51#pullrequestreview-604416303. 

The closure group would like to think about and discuss this. while personally, I think it is more of the responsibility of the FCC group to provide a reasonable solution, as what they did by now is not that interesting (searching for functions missing arguments and fill it back with the default param if possible). Actually, I tried to point out this problem last week https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/pull/15#discussion_r583981098 while it seems that they didn't understand my point at that time.

DDEle:
We fix the conflict (regarding destructure & string?) last night and passed all the test cases. However, we are not sure how to resolve conflicts now (after fancy-calling-conventions (FCC) is merged) as there is a nontrivial design decision about how type-checking functions work if they have default parameters and more problems derived from that in terms of `ea` and `codegen`, as mention by @jpolitz in https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/pull/51#pullrequestreview-604416303. 

The closure group would like to think about and discuss this. while personally, I think it is more of the responsibility of the FCC group to provide a reasonable solution, as what they did by now is not that interesting (searching for functions missing arguments and fill it back with the default param if possible). Actually, I tried to point out this problem last week https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/pull/15#discussion_r583981098 while it seems that they didn't understand my point at that time.

jpolitz:
I'm trying to work through this merge now. Down to the main problems being that e.g. bracket-lookup isn't defined in `ea.ts`. Looking into a few other issues. The main fix is to change `Type` to `Parameter` in many callable places.

jpolitz:
I left some tests with `it.skip`. The ones in `ea` are unmaintainable tests, we need a different way to write those.

We don't understand how lambda type-checking is working right now, and that needs fresh eyes, because I don't think it was working.

DDEle:
Thanks for the work done for merging. Our group distributed work based on the pipeline and @xih108  was in charge of parsing and tc. (BTW we have realized that we should familiar ourself more with other code to tackle the "exam" and “interview” you mentioned yesterday).

I was thought that Python does not support typed parameters in lambda expressions, which probably because that I checked the Python specification (https://docs.python.org/3/reference/expressions.html#lambda) of the early version. Base on that, we thought that we will leave lambdas (which is included in the first 2 test cases on our proposal) until we have some inspirations from the type inference group. I'm sorry for misleading my group members.

I just browse the specification again when writing this comment and found that typed parameters are actually available for lambda expressions, it just doesn't support typing for the expected return value. In this way, we will make some big changes to the lambda expression AST in the following week.

DDEle:
> @jpolitz Thanks for the work done for merging. Our group distributed work based on the pipeline and @xih108 was in charge of parsing and tc. (BTW we have realized that we should familiarize ourselves more with each other's code to tackle the "exam" and “interview” you mentioned yesterday).
> 
> I was thought that Python does not support typed parameters in lambda expressions, which probably because that I checked the Python specification (https://docs.python.org/3/reference/expressions.html#lambda) of the early version. Base on that, we thought that we will leave lambdas (which is included in the first 2 test cases on our proposal) until we have some inspirations from the type inference group. I'm sorry for misleading my group members.
> 
> I just browse the specification again when writing this comment and found that typed parameters are actually available for lambda expressions, it just doesn't support typing for the expected return value. In this way, we will make some big changes to the lambda expression AST in the following week.

It seems that lambda expressions do not support typed param. So the original decision holds. We will try to infer the type of arguments.

From the specifications linked above: 
> See section Function definitions for the syntax of parameter lists. **Note that functions created with lambda expressions cannot contain statements or annotations.**

jpolitz:
Maybe just skip lambda for now? Focus on making the rest clean and complete?

type-check.ts
```diff
@@ -246,6 +405,12 @@ function tcDestructure(
     let { target, starred, ignore } = aTarget;
     const tTarget = tcAssignable(env, locals, target);
     const targetType = tTarget.a;
+
+    if (expr.tag === "lambda") {
+      tcLambda(locals, expr, targetType);
+      valueType = tcExpr(env, locals, expr).a;
+    }
+
```
R167, line 413: Can you say why this actually needed to be hard coded into `tcDestructure` rather than passing in the value type itself??

PR #63, opened 2021-03-05
Testing Milestone

ptliu:
Testing milestone


ptliu:
Linter currently fails on cypress/integration/frontend.spec.js because the linter doesn't know about global variables in Cypress. I believe this is a place we can safely ignore the linter

R167:
> EDIT: we fixed the linter by adding a plugin for cypress.

cool. I was about to say maybe something like that exists, otherwise just ignore that rule for those files. Glad you found a plugin for it!

.eslintrc
```diff
@@ -13,7 +13,8 @@
     "eslint:recommended",
     "plugin:@typescript-eslint/recommended",
     "plugin:prettier/recommended",
-    "prettier"
+    "prettier",
+    "plugin:cypress/recommended"
```
R167, line 17: @edwjchen because of how prettier works to turn off certain rules, it should actually be the last plugin that gets defined, so it should instead be: 
```suggestion
    "plugin:cypress/recommended"
    // Do not add plugins below this line. Prettier documents it must be loaded last.
    "plugin:prettier/recommended",
    "prettier"
```
(unfortunately I can't make the suggestion go across all lines because of a limitation of github).

R167: note: the suggestion only documents what should be done and cannot be merged as is.

.github/workflows/cypress.yml
```diff
@@ -0,0 +1,77 @@
...
+            ${{ runner.os }}-node-
+      # Install using ONLY the package-lock.json
+      - name: Install dependencies
+        run: npm ci
+      - name: Start testing
+        uses: cypress-io/github-action@v2
+        with:
+          start: npm start
+          wait-on: 'http://localhost:8080'
+          command: npm run cypress:run
```
R167, line 75: Any reason this isn't being added as just another job on the existing CI workflow in `.github/workflows/main.yml`? seems like you just copy pasted everything from there.

ptliu: to have the option to turn it off if need be, wanted to preserve the original workflow if something was wrong with this one. Also separates errors from cypress from other errors.

R167: That's fine. In that case, can you remove the duplicate `lint` and `test` runs from this file? Also rename the workflow to something like `Cypress CI`. Otherwise the name shows up the same in Actions and it becomes somewhat confusing.

PR #65, opened 2021-03-05
Milestone Strings

suhas-goutham:
This implementation contains storing and printing strings, and indexing strings(supports negative index too).

### DIFF
--- a/compiler.ts
+++ b/compiler.ts
@@ -75,6 +75,7 @@ export function compile(ast: Program<Type>, env: GlobalEnv): CompileResult {
   definedVars.add("$string_val"); //needed for string operations
   definedVars.add("$string_class"); //needed for strings in class
   definedVars.add("$string_index"); //needed for string index check out of bounds
+  definedVars.add("$string_address"); //needed for string indexing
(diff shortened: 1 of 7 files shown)


suhas-goutham:
Made changes suggested in the previous PR(removed ASCII 0 at the end of each string since we have length at the beginning, negative indexing is now supported)

compiler.ts
```diff
@@ -399,43 +401,39 @@ function codeGenExpr(expr: Expr<Type>, env: GlobalEnv): Array<string> {
         var brObjStmts = codeGenExpr(expr.obj, env);
         var brKeyStmts = codeGenExpr(expr.key, env);
         var brStmts = [];
-        //First check whether index is out of bounds, if so, throw error by going to print_str. Then index the string.
         brStmts.push(
           ...[
             `${brObjStmts.join("\n")}`, //Load the string object to be indexed
+            `(local.set $$string_address)`,
```
jpolitz, line 407: What happens if `brKeyStmts` was compiled from an expression that also compiles to set `$$string_address`?

For example `a[f(b[5])]`?

PR #68, opened 2021-03-07
Fix after Thu/Fir merging from Closure group 

DDEle:
Here are some changes we made after the big merging that Joe made on Thu and Fri. Basically, we (@YaosenLin, @xih108 and me) un-`skip` all the test cases except those ones about lambda expression (as discussed here https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/pull/62#issuecomment-791768998).

I noticed that there are some conflicts now and I'll try to fix them asap after publishing this PR.

We also add more test cases, including all cases in our original proposal except those with lambda expressions. As we carefully designed the structure of escape analysis and the organization of the compiler in the previous week, we managed to pass all of them with a reasonable amount of changes.

This leads to the problem that what is our new plan for 11th Mar (we said that we will complete 2 more cases in our milestone document). Basically, there are 3 options that we can not decide. So we'd like to talk about them here and @jpolitz can advise:
1. Implement the lambda expression as planned in our original proposal. As lambda expressions are not compatible with type annotated parameters, we have to infer the type of them. After discussion, we came up with two ideas to implement that:
    - We can modify the `tcExpr` function to something with an `expected_type` parameter. The drawback is that all usages of `tcExpr` need to be changed, which will confuse other groups working on the *tc*.
    - According to the proposal of the type inference group, another option is to first mark the type of/inside lambda expressions as "TBD" and add an additional step of type inference. However, we are worried that it will be a huge pain to merge our `infer_lambda` stage with that from the type inference group. (It also seems stupid to pipeline both inference stages. 

2.  Working with the keyword and default arguments is another option. I did some study about [CPython's implementation](https://eli.thegreenplace.net/2012/03/23/python-internals-how-callables-work) and realized that probably a subversion is needed to support first-class functions with default/keyword arguments. A randomly proposed structure of function pointer metadata is attached below.
    > The value of a function should be an `i32` pointer pointing to a heap address, which is the start of 7 parts:
    > 1. 1 word. The index of a corresponding function in the table
    > 2. 1 word of value `m`. The number of nonlocal variables used in this function.
    > 3. `m` words. `m` nonlocal variables. 
    > 4. 1 word of value `n`. The number of parameters with no default value
    > 5. `n` words. The pointer of string literals for `n` non-default parameters. (to support arguments that are both positional and keyword-support)
    > 6. 1 word of value `k`. The number of paraments with a default value.
    > 7. `2k` words. `k` tuples of the pointer of string literals of parameter name and the default value
    > 
    > There is probably something else needed (maybe 2 booleans) to support `def foo(*args, **kwargs)`
    
    Which seems far more complicated for a week. 

3. Actually, our `ea` just flatten all functions into closures by now so our third choice is to do some escape analysis and treat some functions as normal wasm function if possible. However, the problem is that this task is not that encouraging.  It will be very hard to demonstrate what we will have done comparing to implementing other features. In other words, we will make a great effort just to pass the same tests we passed now. Besides, it seems difficult for 3 people to work on this concurrently.

    BTW (kind of joking), literally, we are assigned to the feature of *closures, first-class function, and lambda*, while escape analysis is kind of an optimization step.

4. Additional, I guess that it might be a fairly amount of work to do to solve the conflicts about `ea.ts` if there are more incoming merges of other group's PR.


jpolitz:
I wouldn't try to add more. I would make sure to _test_ more, with what you have, and keep it stable. That will be a big accomplishment.

jpolitz:
I think I'm done with my pass over the initial milestones now, so if you clean this up, I can merge it next.

jpolitz (approved):
Generally LGTM, merging to make progress. What's `myMemForward`?

YaosenLin:
> Generally LGTM, merging to make progress. What's `myMemForward`?

It is originally used to update the heap pointer after allocating a bunch of memory. The update of the heap pointer is now bound to the allocating process.

PR #70, opened 2021-03-08
Fix for loop issues & add test cases to the test folder

Quincy0v0:


jpolitz:
Did some more merging, this one is next – can you check on the conflict? I'm not sure I understand `Code_step` vs `Code_step_expr`.

Quincy0v0:
It seems that the AST was changed once again since yesterday. Now our code with `field-assign` is no longer working

Quincy0v0:
Hello prof Joe, 
It seems that a `location` field has been added to the `a` field of the AST. We've fixed our part to be compatible to it, but some tests are failing. All our For loop tests have passed though

R167:
> It seems that the AST was changed once again since yesterday. Now our code with `field-assign` is no longer working

`field-assign` and `bracket-assign` should have been fully removed in anything since #37. I would think any merge resolutions which reintroduced it would be considered unintended since code gen for that form is not supported (everything should be through `assignment` now).

Quincy0v0:
I believe 

> > It seems that the AST was changed once again since yesterday. Now our code with `field-assign` is no longer working
> 
> `field-assign` and `bracket-assign` should have been fully removed in anything since #37. I would think any merge resolutions which reintroduced it would be considered unintended since code gen for that form is not supported (everything should be through `assignment` now).

I see. So we should just remove our `field-assign` case and replace it with `assignment`. Let me try if this would fix it.

R167:
Yeah. Looks like the issue was with the conflict resolution and somethings getting re-added in https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/commit/62ddd160ba4a2a5b9c9c5acc0e2c6ecf6c27bbff. Additionally, since the original PR had merge conflicts, GitHub was unable to run the test action on #58. Since the resolution was done outside of GitHub and merged directly into main, automatic tests were never run for the changes proposed until _after_ it hit main.

Quincy0v0:
Now I fully understand what happened to our code. In #37 the `field-assign` was removed, and another group modified our code to fit this change. However, we haven't uploaded our test cases at that time, so they passed the test even though they broke our code.

R167:
Yup, so a combination of adding back in dead code caused some test cases to continue passing, even though
 the functionality was broken because we didn't have sufficient test coverage to test all the way through. fun times!

Quincy0v0:
> Yup, so a combination of adding back in dead code caused some test cases to continue passing, even though the functionality was broken because we didn't have sufficient test coverage to test all the way through. fun times!

Is there an example how I should construct the `destructure` field in `assignment` if I want to do a field assign? I am having some trouble understanding the code. Thanks!

R167:
You might look at this `utils.test.ts` helper function. tbh, you could probably just move this over to `utils.ts` since I can see it having broader usage than just for tests (or just a similar copy). Either way, better variable names here... :sigh:

https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/blob/dfaa44ba130188c5427fbb8e7c4405c91dd3ffc5/tests/utils.test.ts#L84-L111

There's also `traverseDestructure` which is used to parse code, but that's probably overkill for you here.

https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/blob/dfaa44ba130188c5427fbb8e7c4405c91dd3ffc5/parser.ts#L477

Quincy0v0:
> You might look at this `utils.test.ts` helper function. tbh, you could probably just move this over to `utils.ts` since I can see it having broader usage than just for tests (or just a similar copy). Either way, better variable names here... :sigh:
> 
> https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/blob/dfaa44ba130188c5427fbb8e7c4405c91dd3ffc5/tests/utils.test.ts#L84-L111
> 
> There's also `traverseDestructure` which is used to parse code, but that's probably overkill for you here.
> 
> https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/blob/dfaa44ba130188c5427fbb8e7c4405c91dd3ffc5/parser.ts#L477

Sorry, but I am still a little confused. Say I want to do `rg.cur = 1`, how should I assign the parameters of this function? 
My current understanding is 
{ name: "cur", value: 1,  loc1: `NUM`,  loc2: `rg class`,  loc3: ??? }

Quincy0v0:
> You might look at this `utils.test.ts` helper function. tbh, you could probably just move this over to `utils.ts` since I can see it having broader usage than just for tests (or just a similar copy). Either way, better variable names here... :sigh:
> 
> https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/blob/dfaa44ba130188c5427fbb8e7c4405c91dd3ffc5/tests/utils.test.ts#L84-L111
> 
> There's also `traverseDestructure` which is used to parse code, but that's probably overkill for you here.
> 
> https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/blob/dfaa44ba130188c5427fbb8e7c4405c91dd3ffc5/parser.ts#L477

I am looking at `codeGenDestructure`. It seems that we can only assign all class

Quincy0v0:
> You might look at this `utils.test.ts` helper function. tbh, you could probably just move this over to `utils.ts` since I can see it having broader usage than just for tests (or just a similar copy). Either way, better variable names here... :sigh:
> 
> https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/blob/dfaa44ba130188c5427fbb8e7c4405c91dd3ffc5/tests/utils.test.ts#L84-L111
> 
> There's also `traverseDestructure` which is used to parse code, but that's probably overkill for you here.
> 
> https://github.com/ucsd-cse231-w21/chocopy-wasm-compiler/blob/dfaa44ba130188c5427fbb8e7c4405c91dd3ffc5/parser.ts#L477

I am looking into `codeGenDestructure`. It seems that I have to assign all fields in a class at a time. Is that correct?

R167:
> Sorry, but I am still a little confused. Say I want to do `rg.cur = 1`, how should I assign the parameters of this function?
> My current understanding is
> { name: "cur", value: 1, loc1: `NUM`, loc2: `rg class`, loc3: ??? }

yeah, so rn that function only does `id` based assign.

You would want something like
```typescript
const obj = /* Expr for rg */;
const field = "cur";
const value = /* value Expr */;
const dummyLocation = /* location */;
const assign = {
  a: dummyLocation,
  destruct: {
    isDestructured: false,
    targets: [
      {
        ignore: false,
        starred: false,
        target: {
          a: dummyLocation,
          tag: "lookup",
          field: field,
          obj: obj,
        },
      },
    ],
    valueType: dummyLocation,
  },
  tag: "assignment",
  value: value,
}
```

Quincy0v0:
> > Sorry, but I am still a little confused. Say I want to do `rg.cur = 1`, how should I assign the parameters of this function?
> > My current understanding is
> > { name: "cur", value: 1, loc1: `NUM`, loc2: `rg class`, loc3: ??? }
> 
> yeah, so rn that function only does `id` based assign.
> 
> You would want something like
> 
> ```ts
> const obj = /* Expr for rg */;
> const field = "cur";
> const value = /* value Expr */;
> const dummyLocation = /* location */;
> const assign = {
>   a: dummyLocation,
>   destruct: {
>     isDestructured: false,
>     targets: [
>       {
>         ignore: false,
>         starred: false,
>         target: {
>           a: dummyLocation,
>           tag: "lookup",
>           field: field,
>           obj: obj,
>         },
>       },
>     ],
>     valueType: dummyLocation,
>   },
>   tag: "assignment",
>   value: value,
> }
> ```

So the `location` doesn't actually matter, right?

R167:
> So the `location` doesn't actually matter, right?
Location is only being used for error reporting. Maybe set to an existing location for the for loop's value and target so if there is an issue, e.g. in type checking, a relevant line is displayed

R167:
note: long term, may also make sense having `for` assign to a destruct rather than using a `name: string` forcing single id only(?) (e.g. make `for c.x in range(1, 5): print(c.x)` valid)

Quincy0v0:
All bugs fixed & all tests passed. @R167 Special Thanks!

jpolitz (approved):

PR #75, opened 2021-03-10
Testing automatic fuzzing and testing final submission

ptliu:
Contributions with automatic fuzizng


ptliu:
Progress: Finished classes, set up automatic fuzzing server with webhook.

New example generated program:
class class_0(object):
  number_0:int = 1
bool_0:bool = False
bool_1:bool = False
number_0:int = 1

class_0_0:class_0 = None
class_0_1:class_0 = None
if False:
  1
  bool_0 = True
  number_0
  if False:
    bool_1
  1
elif (not False) :
  bool_0 = (-(-1)) > number_0 + 1 % 1 // number_0
else:
  1
class_0_0 = class_0_1

edwjchen:
Posting example of a more interesting class generation example: 
- includes methods using keyword self
- includes constructor calls

class class_0(object):
  pass
class class_0(object):
  number_0:int = 1
  def func_number_0(self: class_0, func_number_0_param_0: int) -> int:
    class_0_0.func_number_0(1)
    None
    return self.number_0 // self.number_0
class_0_0:class_0 = None
class_0_0 = class_0()
print(class_0_0)

ptliu:
Worth noting that this example (other than not being indented) is not legal ChocoPy due to redeclaring a class. This is unfortunately unavoidable in the example we feed to python-shell due to the python-shell bug mentioned earlier.

jpolitz (approved):
LGTM

PR #80, opened 2021-03-10
Fix bugs in print index & change `name` into a Destructure

Quincy0v0:


Quincy0v0:
@student-of It seems that your team added some wrapper to our for loop code in compiler in a previous PR. Can you help us resolve the conflict? Cuz we don't quite understand what the added code is doing here. Thanks!

parser.ts
```diff
@@ -695,14 +699,25 @@ export function traverseStmt(c: TreeCursor, s: string): Stmt<Location> {
...
         c.nextSibling(); // Focus on var name
         name = s.substring(c.from, c.to);
+        ass = { tag: "id", name: name };
+        targets.push({
+          target: ass,
+          starred: false,
+          ignore: false,
+        });
         c.nextSibling(); // Focus on in
       }
```
thecooltechguy, line 722: Hi @Quincy0v0 , in this section, does this mean that the for loop doesn't support arbitrary de-structuring (across `n` targets), and instead only supports de-structuring up to `2` targets (which I'm assuming is cases like `index, name` for `enumerate`)?

Quincy0v0: Ahh this is a thing. I would just change that into a `while` loop so I could parse all targets. That would be an easy fix though

thecooltechguy: This would be the right approach, right @R167 ?

thecooltechguy: That sounds great -- definitely makes sense. Do you know which team's PR would implement this?

R167: I would say you do it since you need it? Honestly, I might just make it a required value rather than optional. Currently only being used in `traverseAssignment`. The only other people I could think might need it would be comprehensions.

Quincy0v0: I've changed the parser and it should work now!

R167:
Heads up, you really might want to resolve that merge conflict so CI can start running again. If memory serves, it's probably because of the changes to memory management.

Quincy0v0:
> Heads up, you really might want to resolve that merge conflict so CI can start running again. If memory serves, it's probably because of the changes to memory management.

I'm not quite sure what the added memory code is doing, so I am waiting for someone from the memory team.

thecooltechguy:
@Quincy0v0 Also, are `for` loops only planning to support `id`-based `Assignable`s, and not the other types of `Assignable`s (e.g., `lookup`, `bracket-lookup`)?

Just asking so that for list comprehensions, we can make sure that we have the same type of functionality/limitations.

parser.ts
R167: > I've changed the parser and it should work now!

Some reason i'm not seeing these changes to `traverseDestructure` in the diff(?)

Quincy0v0:
> @Quincy0v0 Also, are `for` loops only planning to support `id`-based `Assignable`s, and not the other types of `Assignable`s (e.g., `lookup`, `bracket-lookup`)?
> 
> Just asking so that for list comprehensions, we can make sure that we have the same type of functionality/limitations.

I may need to discuss with my team to confirm on that. We don't know if we would have time to handle this. 
BTW, I've resolved the conflict & passed all checks. This should be ready to merge

compiler.ts
```diff
@@ -312,145 +321,302 @@ function codeGenStmt(stmt: Stmt<[Type, Location]>, env: GlobalEnv): Array<string
...
+            (block
+              (loop
+                ${Code_idstep.join("\n")}
+                (br_if 1 ${Code_cond.join("\n")} ${decodeLiteral.join("\n")})
+
+                ${Code_ass.join("\n")}
+                ${bodyStmts.join("\n")}
+                (br 0)
+            ))`,
+          ];
```
student-of, line 453: Missed a temp. guard and join removal here

Quincy0v0: Fixed. Does it look right now?

Quincy0v0:
> @Quincy0v0 I agree it doesn't seem like we would have to time to handle these either. So it's probably safest to just support `id`-based `Assignable`s for now.

Let's just support id-based assignables for now.

compiler.ts
student-of: Looks good! Thanks

Quincy0v0:
Hello Professor Joe, 
This should be ready to merge
@jpolitz

ast.ts
```diff
@@ -76,7 +76,7 @@ export type Stmt<A> =
   | { a?: A; tag: "field-assign"; obj: Expr<A>; field: string; value: Expr<A> }
   | { a?: A; tag: "continue"; depth?: number }
   | { a?: A; tag: "break"; depth?: number } // depth is used for wasm 'br' instruction
-  | { a?: A; tag: "for"; name: string; index?: string; iterable: Expr<A>; body: Array<Stmt<A>> }
+  | { a?: A; tag: "for"; id: number; name: Destructure<A>; iterable: Expr<A>; body: Array<Stmt<A>> }
```
thecooltechguy, line 79: @Quincy0v0 Is there a reason why we still need `id: number`, since any indices returned via `enumerate` would just be one of the `name: Destructure<A>`'s `targets`, right? Or am I misunderstanding something?

thecooltechguy:
@Quincy0v0 In this PR, there doesn't seem to be any tests for multiple de-structured fields and for loops iterating through lists. Are these supported by the current PR? If so, would it be possible to include some tests for these cases?

ast.ts
Quincy0v0: This is not the same `id` here. This `id` is used to help us implement multiple for loops, since we are implementing for loops as objects and each for object requires an `id` to be properly saved.

Quincy0v0:
> @Quincy0v0 In this PR, there doesn't seem to be any tests for multiple de-structured fields and for loops iterating through lists. Are these supported by the current PR? If so, would it be possible to include some tests for these cases?

we don't support multiple de-structured fields in this PR. For lists, we currently only support in-line lists such as `for i in [1,2,3]:`. We plan to expand this, and also support `enumerate` later today.

thecooltechguy:
@Quincy0v0 @student-of I tried merging this PR's main branch into our comprehensions branch at our repo, and now, if we run the following program:

```python
i : int = 0
```

we get the following error:
```
TypeError: Cannot read property 'staticAlloc' of undefined
```

Is this being caused due to some incorrect merging of the earlier merge conflict?

Quincy0v0:
> ```python
> i : int = 0
> ```

This in strange. I never changed any code other than our for loop's code. I tried to run this program, and actually if you click "run" only once it will run successfully. But if you click "run" again you will get this error

thecooltechguy:
> > ```python
> > i : int = 0
> > ```
> 
> This in strange. I never changed any code other than our for loop's code. I tried to run this program, and actually if you click "run" only once it will run successfully. But if you click "run" again you will get this error

Yep you're right, the issue actually only happens if we "run" more than once, definitely strange.

Quincy0v0:
> > > ```python
> > > i : int = 0
> > > ```
> > 
> > 
> > This in strange. I never changed any code other than our for loop's code. I tried to run this program, and actually if you click "run" only once it will run successfully. But if you click "run" again you will get this error
> 
> Yep you're right, the issue actually only happens if we "run" more than once, definitely strange.
> 
> It will occur even if you "run", "reset", and then "run" again. The error will only go away if we refresh the page.

But I don't think this bug is related to the for loop though, since we don't even have a for loop here

thecooltechguy:
@Quincy0v0 You're right, so I will just record another issue for this, thanks for confirming.

thecooltechguy:
There is actually an error tied to this PR:

If we run this in the web ui:
```
for i in range(10):
  print(i)
```

I get the following error:
```
run failed InternalExpcetion: Report this as a bug to the compiler developer, this shouldn't happen 
    at envLookup (http://localhost:8080/webstart.js?def3b14567ebd2a5207a:24608:15)
    at codeGenExpr (http://localhost:8080/webstart.js?def3b14567ebd2a5207a:25319:39)
    at codeGenExpr (http://localhost:8080/webstart.js?def3b14567ebd2a5207a:25285:30)
    at codeGenStmt (http://localhost:8080/webstart.js?def3b14567ebd2a5207a:24628:29)
    at http://localhost:8080/webstart.js?def3b14567ebd2a5207a:24650:58
    at Array.map (<anonymous>)
    at codeGenStmt (http://localhost:8080/webstart.js?def3b14567ebd2a5207a:24650:39)
    at http://localhost:8080/webstart.js?def3b14567ebd2a5207a:24580:51
    at Array.map (<anonymous>)
    at Module.compile (http://localhost:8080/webstart.js?def3b14567ebd2a5207a:24580:37)
```

The program is invalid since I didn't do a `VarInit` for `i`, but shouldn't this throw some other type of error? I was assuming that `tcDestructure` would catch this

Quincy0v0:
> I think there may be an error tied to this PR:
> 
> If we run this in the web ui:
> 
> ```
> for i in range(10):
>   print(i)
> ```
> 
> I get the following error:
> 
> ```
> run failed InternalExpcetion: Report this as a bug to the compiler developer, this shouldn't happen 
>     at envLookup (http://localhost:8080/webstart.js?def3b14567ebd2a5207a:24608:15)
>     at codeGenExpr (http://localhost:8080/webstart.js?def3b14567ebd2a5207a:25319:39)
>     at codeGenExpr (http://localhost:8080/webstart.js?def3b14567ebd2a5207a:25285:30)
>     at codeGenStmt (http://localhost:8080/webstart.js?def3b14567ebd2a5207a:24628:29)
>     at http://localhost:8080/webstart.js?def3b14567ebd2a5207a:24650:58
>     at Array.map (<anonymous>)
>     at codeGenStmt (http://localhost:8080/webstart.js?def3b14567ebd2a5207a:24650:39)
>     at http://localhost:8080/webstart.js?def3b14567ebd2a5207a:24580:51
>     at Array.map (<anonymous>)
>     at Module.compile (http://localhost:8080/webstart.js?def3b14567ebd2a5207a:24580:37)
> ```
> 
> The program is invalid since I didn't do a `VarInit` for `i`, but shouldn't this throw some other type of error? I was assuming that `tcDestructure` would catch this.
> 
> The program does work if I add `i : int = 0` at the very beginning. Maybe, in the type-checking for `for` loops, do we have to add an additional check to ensure all de-structured targets have already been defined in the env with the correct type?
> 
> @R167 Is `tcDestructure` supposed to catch this error?

We actually called `tcDestructure` to handle this...

thecooltechguy:
Yeah I saw that the `for` loops type-checking calls `tcDestructure` on the `name`, and we did something similar for `comprehensions`, so this maybe has to do with `tcDestructure` @R167

compiler.ts
```diff
@@ -95,10 +96,18 @@ export function augmentEnv(
...
+    let rgAddr = mm.staticAlloc(4n);
+    newGlobals.set("rng" + index, Number(rgAddr));
+    mm.addGlobal(rgAddr);
+  }
+  // set bars for indexing in for loops
+  for (let index = lastCount + 1; index <= forCount; index++) {
+    let idxAddr = mm.staticAlloc(4n);
+    newGlobals.set("idx" + index, Number(idxAddr));
+    mm.addGlobal(idxAddr);
+  }
```
student-of, line 110: Are these used as pointers in any capacity? Otherwise, they should not be marked as roots. From one of my tests, I'm guessing that they are being used as untagged numbers.

Quincy0v0: What do you mean by pointers here? I don't quite understand. These are variables that save the information of `for loop` objects

student-of: Are they ever used to store pointers to values? Or are they only used as numbers (if they are, are they tagged as primitives?)

As in, does the GC need to scan them to check object reachability?

Quincy0v0: Oh yeah, the first one, `rgAddr` are pointers to `for loop` objects, and the `idxAddr` are numbers.

student-of: @Quincy0v0 Sorry, lack of sleep made me misread that. Just remove `mm.addGlobal(idxAddr)`

PR #83, opened 2021-03-11
Front-end Final Pull Request

shanliyu:
## Working Examples from Proposal (9 out of 11): 
1. Tooltip on printed output (types): 
When the mouse hover over the printed output string, a tooltip will appear and displays the type of the printed output. 

image

2. Media queries for different breakpoints: 
We add breakpoints to change the layouts and sizes of the editor and interactions panels based on the size of the tab. The two panels are resized to fit the tab and are vertically aligned once the window size is smaller than 840px. The sizes of the editor and interactions panels will get larger if the tab is larger than 800px to provide better user experience when their device has a large screen. The change of the size of the CodeMirror editor is done by overriding the .CodeMirror class.

image

3. Underlining where in program errors have occurred (after compiling): 
We worked with the error-reporting team and implemented this feature. From error-reporting’s error callStack, it contains the line numbers and the error messages of the error. So once we grab the line number and error message, we will use that to highlight the line in the code editor with light red background, and also a lint-like red small rectangular next to the line number, which will show the error message once the cursor hovers over. 

image

image

4. Autocomplete basic keywords (like return, class, etc.): 
We were able to autocomplete basic keywords and built-in function names. We defined a list of words to autocomplete based on https://docs.python.org/3/library/functions.html and https://www.programiz.com/python-programming/keyword-list.

image

5. Autocomplete variables/function definitions:
We were also able to autocomplete variable names, class names, top-level functions, and class method names. We used the repl’s environment to grab the respective strings and insert into our list of words that we could potentially autocomplete. Whenever the “enter” button is pressed in the code editor, the compiler compiles the code in the background (doesn’t render errors) such that the autocomplete could have a richer set of words to suggest. We thought this heuristic fits well considering Python uses whitespace indentation and new lines of code are written usually after “enter” is pressed.

image

6. Change style (theme) of the code editor. 
We created a dropdown menu for the user to choose all available editor themes provided by CodeMirror. When a user clicks on the dropdown and chooses the desired theme, the editor will change the theme accordingly. 

image

7. Optimize the appearance of the web page
We used SCSS to style the web page.  HTML elements are manipulated in webstart.js when objects, lists, dicts are printed, so that we have a better presentation of those structures.  The REPLs/Outputs panel can be hidden if the user chooses to enter a full-screen edit mode in the editor. A hotkey of (Ctrl + R) for running the code is added.

image

8. A button for clearing the REPLs and resetting the environment: 
We created a button on the editor side of the webpage. This removes the REPL outputs, creates a new environment and removes any of text in the editor. This functionality is exactly like pressing the “refresh button” on a browser.

image

image

9. Save repl code and load repl code
We added a button to download the chocopy code in the editor with the user selected name to the local file folder and added a button to load the code (with extensions of .txt or .py) from a local file folder into the editor. This enables the user to import and run their code efficiently.

image

image

## Neglected Examples from Proposal(2 out of 11)
1. Highlight syntax error. 
Codemirror does not have such functionality by default, while there is a community project (codemirror.net/1/contrib/python), but the functionality is very minimal, and most of the implementation does not relate to any compiler knowledge we have studied during this class, therefore we decided to neglect this example. 

2. Multiple editor tabs: 
After some review, we think that this functionality is almost identical to having multiple browser tabs open, therefore we have neglected this example for our final implementation. 

## Additional Implemented features not in the proposal (2 new features)
1. A button to just remove REPL outputs
We added a button to just the REPL side of the webpage to remove outputs and any REPL executed code. This would be useful if the user has many outputs or has utilized the REPL many times as the user can declutter their workspace.

image

image

2. Pretty print objects, lists and dictionaries. 
Based on the professor's suggestion, we worked with the list team and dictionary team to implement a feature called pretty printing, which has similar functionality to how Chrome browser developer tool console prints objects. 
When an object, list, or dictionary is being rendered and evaluated, we will have a collapsible which can be expanded to display all the items according to their fields, indices, or keys. 
Below screenshot shows an object with a field named q, which is a list of dictionary with int key and Rat value. On the right shows the pretty printed output with all collapsibles expanded: 

image


jpolitz:
Maybe @ptliu or someone on testing could help us understand the Cypress error?

ptliu:
ah, our test was based on the old frontend, so it's looking for an element that doesn't exist. It seems to me in this case it makes sense to rewrite the test in this case

PR #90, opened 2021-03-11
Memory Management Final Submission

student-of:
Stop-the-world, mark-and-sweep garbage collector enabled by default and guaranteed to run after each REPL entry.

All 10 example programs pass, testing before/after memory usage.
All other tests pass.


jpolitz (approved):
LGTM

PR #94, opened 2021-03-12
Final Submission for Closures

YaosenLin:
- We successfully implement all the examples in the initial proposal except the one with lambda expression.
  1. A trivial nested function without escape  
      ``` python
      def f(x: int) -> int:
        def inc() -> int:
          return x + 1
        return inc()
      f(5)
      ```

  2. A trivial `nonlocal` usage  
      ``` python
      def f(x : int) -> int:
        def g(y : int) -> int:
          return x + h(y) 
        def h(z : int) -> int:
          nonlocal x
          x = z
          return x + 1
        return g(10) + g(7)
      f(6)
      ```

  3. A trivial case where `nonlocal` can be eliminated  
      ``` python
      def f() -> int:
          x: int = 0
          def g() -> int:
              nonlocal x
              return x + 1
          return g()
      f()
      ```

  4. A trivial example of closure from the lecture note (function escapes when it is returned)  
      ``` python
      def f(x : int)->Callable[[int],int]:
        def g(y : int) -> int:
          return x + y
        return g
  
      g_where_x_is_6: Callable[[int],int] = None
      g_where_x_is_6 = f(6)
      print(g_where_x_is_6(5))
      ```

  5. A closure with a variable of object type  
      ``` python
      class A(object):
        x:int = 1
  
      def f(a:A)->Callable[[int],int]:
        def g(y: int) -> int:
          a.x = a.x + y
          return a.x
        return g
  
      a:A = None
      g: Callable[[int], int] = None
      a = A()
      a.x = 6
      g = f(a)
      a.x = 10
      g(2) + a.x
      ```

  6. A modified example from the lecture note (`nonlocal`, `inc`/`dec`) (multiple functions returned)  
      ``` python
      class Triplet(object):
        fst:Callable[[], int] = None
        snd:Callable[[], int] = None
        thd:Callable[[], int] = None
  
  
      def foo() -> Triplet:
        x: int = 0
        r: Triplet = None
        def inc() -> int:
          nonlocal x
          x = x + 1
          return x
        def dec() -> int:
          nonlocal x
          x = x -1
          return x
        def curr() -> int:
          return x
        r = Triplet()
        x = 100
  
        r.fst = inc
        r.snd = dec
        r.thd = curr
        
        return r
      ```

  7. An non-escaping function passed to another function as a callable argument  
      ``` python
      def f(x : int) -> Callable[[int], bool]:
        def g(y : int) -> bool:
          return x > y
        return g
  
      def id(c : Callable[[int], bool]) -> Callable[[int], bool]:
        return c
  
      id(f(10))(5)
      ```

  8. An escaping function passed to another function as a callable argument  
      ``` python
      class MyTupleInt(object):
        fst: int = 0
        snd: int = 0
  
      def add_n(x:int) -> Callable[[int], int]:
        def g(y:int)-> int:
          return x + y
        return g
  
      def map2(x: int, y: int, f: Callable[[int], int])-> MyTupleInt:
        t: MyTupleInt = None
        t = MyTupleInt()
        t.fst = f(x)
        t.snd = f(y)
        return t
  
      add_2: Callable[[int], int] = None
      r: MyTupleInt = None
  
      add_2 = add_n(2)
      r = map2(3, 5, add_2)
      r.fst
      r.snd
      ```

  9. An example from Feb 16 lecture (an escaping function calls its non-escaping sibling)  
      ``` python
      def f(x:int) -> Callable[[], int]:
          def g() -> int:
              return h()
          def h() -> int:
              return x + 1
          return g
  
      f(10)()
      ```

2. More tests are added in `closures.test.ts`, including those for future extending features like nested class method.
3. Supporting `list-expr` in `ea.ts` is still on the way, please **keep the PR open**


tests/closures.test.ts
```diff
@@ -162,4 +162,155 @@ describe("Closures group's test cases on the proposal", () => {
...
+  x = f(x)
+  `;
+  asserts("15. A nested function with dict", [
+    [src, PyNone()],
+    ["x[1]", PyInt(2)],
+    ["x[11]", PyInt(22)],
+    ["x[31]", PyInt(56)],
+  ]);
+
+  describe.skip("nested func - loop", function () {
```
DDEle, line 268: We added some test cases about using other group's features inside functions. Note that there are 2 cases that are marked as skipped now: a case for list and a case for `for`.
For the `list` case, we can not make it run, probably because that we are not that familiar with the strategies used by the list group. The program raised an exception about memory outage but we were not managed to find anything weird on the flattened AST.

tests/closures.test.ts
```diff
@@ -162,4 +162,155 @@ describe("Closures group's test cases on the proposal", () => {
...
+  x = f(x)
+  `;
+  asserts("12. A nested function with list created", [
+    [src, PyNone()],
+    ["x[0]", PyInt(1)],
+    ["x[1]", PyInt(1)],
+    ["x[2]", PyInt(1)],
+  ]);
+
+  describe.skip("nested func - list", function () {
```
DDEle, line 205: We added some test cases about using other group's features inside functions. Note that there are 2 cases that are marked as skipped now: a case for list and a case for for.
For the list case, we can not make it run, probably because that we are not that familiar with the strategies used by the list group. The program raised an exception about memory outage but we were not managed to find anything weird on the flattened AST.

tests/closures.test.ts
DDEle, line 268: Another skipped case is about the `for` loop. We expect some big change to make it work as currently the assignee is a string rather than a `Destructure` , which is an obstacle for wrapping and pass around this variable. If we have time. we can work on this later with an additional PR after the final PR of the for-loop team got merged.

PR #97, opened 2021-03-15
Dictionary Mark-and-Sweep Tracing Fix

student-of:
Dictionaries are arrays of POINTERS to dictionary entries, not an array of dictionary entries.

Add an integration REPL test and a simulated unit test.


jpolitz (approved):
Good catch

PR #110, opened 2022-04-02
Add ir

jpolitz:
Creating a PR to look at diffs


ir-compiler.ts
```diff
@@ -0,0 +1,256 @@
...
+      var className = objTyp.name;
+      var [offset, _] = env.classes.get(className).get(stmt.field);
+      var valStmts = codeGenValue(stmt.value, env);
+      return [
+        ...objStmts,
+        `(i32.add (i32.const ${offset * 4}))`,
+        ...valStmts,
+        `(i32.store)`
+      ];
+
```
jpolitz: Is the plan for the cases below to do more or less a while/if with an integer to simulate basic blocks? Or do something clever like Stackifier?

https://medium.com/leaningtech/solving-the-structured-control-flow-problem-once-and-for-all-5123117b1ee2

jpolitz: One option is to do the switch/loop and have it be a project to improve it!

